PR #2, opened 2016-01-01
Import containers when accessing dependencies via `import_module`

timriley:
I found this bug when working on icelab/alpinist#2 – while the app would work properly when run via the web server, my unit specs were failing because the "core" dependencies (i.e. the ones imported into the `Main::Container` from `Alpinist::Container`) were not being loaded properly. This was because in the unit  specs were (rightly) not finalizing `Main::Container`, in order to test things in isolation and to avoid spuriously loading the whole application environment.

To address this, I've made it so that any imported containers are imported when auto-imported dependencies are accessed from a container's `.import_module`.

_(I'm not 100% sure if this is the approach you'd like to take, @solnic, so I was going to push this up into a branch directly on the dryrb repo so you could potentially make adjustments, but dry-container isn't included in the list of repos I can write to, so I've put it up on a personal fork instead. Let me know if you'd like me to push this anywhere else)._

### DIFF
--- a/lib/dry/component/container.rb
+++ b/lib/dry/component/container.rb
@@ -58,7 +58,7 @@ module Dry
       def self.finalize!(&_block)
         yield(self) if block_given?
 
-        imports.each { |ns, container| import_container(ns, container) }
+        import_containers
 
         Dir[root.join("#{config.core_dir}/boot/**/*.rb")].each do |path|
           boot!(File.basename(path, '.rb').to_sym)
@@ -177,6 +177,10 @@ module Dry
       end
 
       private
+
+      def self.import_containers
+        imports.each { |ns, container| import_container(ns, container) }
+      end
 
       def self.import_container(ns, container)
         container.finalize!
--- a/lib/dry/component/container.rb
+++ b/lib/dry/component/container.rb
@@ -73,6 +73,7 @@ module Dry
         auto_inject = Dry::AutoInject(self)
 
         -> *keys {
+          import_containers
           keys.each { |key| load_component(key) unless key?(key) }
           auto_inject[*keys]
         }
--- a/lib/dry/component/container.rb
+++ b/lib/dry/component/container.rb
@@ -176,13 +176,23 @@ module Dry
         @imports ||= {}
       end
 
+      def self.imported
+        @imported ||= {}
+      end
+
       private
+
+      def self.imported?(name)
+        imported.key?(name)
+      end
 
       def self.import_containers
         imports.each { |ns, container| import_container(ns, container) }
       end
 
       def self.import_container(ns, container)
+        return if imported?(ns)
+
         container.finalize!
 
         items = container._container.each_with_object({}) { |(key, item), res|
@@ -190,6 +200,8 @@ module Dry
         }
 
         _container.update(items)
+
+        imported[ns] = true
       end
 
       def self.auto_register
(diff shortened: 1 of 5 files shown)


lib/dry/component/container.rb
```diff
@@ -176,9 +177,23 @@ def self.imports
...
+
       private
 
+      def self.imported?(name)
+        imported.key?(name)
+      end
+
+      def self.import_containers
+        imports.each { |ns, container| import_container(ns, container) }
+      end
```
solnic: `map` would be more idiomatic here, I guess :)

solnic: Oh wait. Nevermind. The name of the meth confused me. Maybe should be sth like `load_import_containers` or something

solnic:
There's one tweak I though about - skip import containers finalization when deps are being dynamically resolved w/o complete finalization. Otherwise loading a component might cause other things to be booted etc.

lib/dry/component/container.rb
AMHOL: Think `load_imported_containers` works as far as I can tell

AMHOL: Ignore me lol, shouldn't make comments when I only have my phone to access github =P

timriley:
Thanks for the feedback! I'll go about separating the loading and finalizing of the imported containers, and improving overall naming.

timriley:
So I've just tweaked the method names and added support for not finalizing the imported containers if we're just resolving individual deps. I haven't committed yet (I'll explain in a moment), but for reference, the code looks like this:

```ruby
def self.load_imported_containers(finalize: false)
  imports.each { |ns, container| load_imported_container(ns, container, finalize: finalize) }
end

def self.load_imported_container(ns, container, finalize: false)
  return if imported?(ns)

  container.finalize! if finalize

  items = container._container.each_with_object({}) { |(key, item), res|
    res[[ns, key].join(config.namespace_separator)] = item
  }

  _container.update(items)

  imported[ns] = true
end
```

Code-wise, I thought the named `finalize` kwarg made it clear at the call-site what we'd be expecting the method to do re: finalizing or not, and it seemed like it'd be a worse solution to have a separate methods for loading vs loading+finalizing.

Anyway, here's the problem I hit: in the unit test (which hasn't changed) my imported container has auto-registration enabled, and the files to be loaded via auto-registration don't get loaded until that container is actually finalized. So when I have this inside `.import_module`...

```ruby
load_imported_containers(finalize: false)
```

...it doesn't actually make my auto-registered deps available from that imported module, which means the test fails.

This makes me think that we might actually be better served by a different approach.

Perhaps instead we have `.require_component` do this:

1. Try to require the component from the current container (i.e. the current behaviour)
2. If this finds nothing (i.e. before we `fail ArgumentError` atm), loop through any yet-to-be-imported containers for the component (taking away the imported container's local name prefix) and try and `.require_component` the dep there.

This way we could load the dep from the other container and avoid finalizing it, which is our preferred behaviour, and we can get rid of the "brute-force" bulk import of all the containers in `.import_module`. Seems like it'd be cleaner and more direct.

Does that seems like a reasonable approach? If you think you'd be happy with this, I'll put it together for you.

solnic:
@timriley this sounds good, in fact that's what I wanted to do initially but then, being tired and sleepy (or maybe just hungry?) I ended up doing it differently. But what you describe makes sense. I struggled with test suite (loaded_features issue) so I rushed with it too quickly, I suppose :/

timriley:
Closing this since #3 is implementing the alternative approach I suggested in the final comment (I hope to loop back to #3 and get it merged at some point soon).

PR #3, opened 2016-01-04
Require components from imported containers

timriley:
This implements the approach I suggested at the end of #2:

> Perhaps instead we have .require_component do this:
>
> 1. Try to require the component from the current container (i.e. the current behaviour)
> 2. If this finds nothing (i.e. before we `fail ArgumentError` atm), loop through any yet-to-be-imported containers for the component (taking away the imported container's local name prefix) and try and `.require_component` the dep there.
>
> This way we could load the dep from the other container and avoid finalizing it, which is our preferred behaviour, and we can get rid of the "brute-force" bulk import of all the containers in .import_module. Seems like it'd be cleaner and more direct.

This looks like it works nicely. It gives us dependencies from imported containers in isolation, and allows us to avoid trying to import all the containers wholesale whenever we're auto-injecting a dependency of any kind.

If you're happy with this, we can close #2.

### DIFF
--- a/lib/dry/component/container.rb
+++ b/lib/dry/component/container.rb
@@ -143,6 +143,14 @@ module Dry
           Kernel.require component.path
           yield(component.constant) if block
         else
+          imports.each do |ns, container|
+            begin
+              localized_key = key.sub(/^#{Regexp.escape(ns)}#{Regexp.escape(config.namespace_separator)}/, '')
+              return container.require_component(localized_key, &block)
+            rescue ArgumentError
+            end
+          end
+
           fail ArgumentError, "could not resolve require file for #{key}"
         end
       end
--- a/spec/fixtures/import/lib/test/import_dep.rb
+++ b/spec/fixtures/import/lib/test/import_dep.rb
@@ -1,0 +1,4 @@
+module Test
+  class ImportDep
+  end
+end
--- a/spec/fixtures/test/lib/test/using_import.rb
+++ b/spec/fixtures/test/lib/test/using_import.rb
@@ -1,0 +1,5 @@
+module Test
+  class UsingImport
+    include Import['other.test.import_dep']
+  end
+end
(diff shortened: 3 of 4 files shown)


lib/dry/component/container.rb
```diff
@@ -143,6 +143,14 @@ def self.require_component(key, &block)
...
           yield(component.constant) if block
         else
+          imports.each do |ns, container|
+            begin
+              localized_key = key.sub(/^#{Regexp.escape(ns)}#{Regexp.escape(config.namespace_separator)}/, '')
+              return container.require_component(localized_key, &block)
+            rescue ArgumentError
+            end
+          end
+
```
timriley: The only code/style question I have for this PR is whether you might like this `imports.each` block moved into a private method for succinctness/clarity of the main `.require_component` method. I thought about it briefly, but I thought there was a chance it might actually harm readability/understandability, since it obscures the early return that we need to get out of the method before that final `fail`.

timriley:
Don't forget this PR :) It's ready to go, pending your feedback.

lib/dry/component/container.rb
solnic: This looks like we're missing some API. Would it be possible to ask a container whether a dependency exists or not?

solnic:
@timriley I'm sorry mate, I missed this one...you know how it goes!

solnic:
@timriley I had [a local commit](https://github.com/dryrb/dry-component/commit/2e6bde237b91db32cc48854a6d5ca1d1dada22fd) that I forgot to push...not sure if that addresses the same thing or not, I *forgot* what I did there and why :joy: could you see if this is working for you? (it's in master now)

timriley:
Yeah, I'm pretty sure this commit of yours (from back in February, haha!) did what I was looking to achieve with this PR. Closing!

PR #4, opened 2016-01-28
Multiple config files

sagmor:
There's currently an error where you can't have multiple `yml` config files for different components.
Because `Dry::Component::Config` would get "locked" with the first loaded file's keys.


solnic:
Oh nice catch, thank you!

PR #6, opened 2016-02-10
Namespaces proof of concept

sagmor:
An issue I'm getting with components is that I like to keep my code inside a single `module` which leads to file structures like:

```ruby
# To auto register operations inside:
#   my_app/components/sub_app/lib/my_app/sub_app/operations/**/*.rb
module MyApp::SubApp
  class Container < Dry::Component::Container
    configure do |config|
      config.root = ROOT.join('my_app/components/sub_app')
      config.auto_register = 'lib/my_app/sub_app/operations'
    end

    load_paths! 'lib'
  end
end

MyApp::SubApp::Container.finalize!
MyApp::SubApp::Container['my_app.sub_app.operations.some_operation']
```

This PR is a bare implementation of a Namespaced component.
The idea is that the component "knows" everything livies inside a module so it loads everything inside it.
So for example the same scenario from above would look more like:

```ruby
# To auto register all operations, now inside:
#   my_app/components/sub_app/operations/**/*.rb
module MyApp::SubApp
  extend Dry::Component::Namespace
  configure do |config|
    config.root = ROOT.join('my_app/components/sub_app')
    config.auto_register = 'operations'
  end
end

MyApp::SubApp.finalize!
MyApp::SubApp::Container['operations.some_operation']
```

Just wanted to see what you think about this concept before investing more time on the idea :wink:

### DIFF
--- a/lib/dry/component/container.rb
+++ b/lib/dry/component/container.rb
@@ -34,6 +34,10 @@ module Dry
         self
       end
 
+      def self.Loader(key)
+        Component.Loader(key)
+      end
+
       def self.import(other)
         case other
         when Dry::Container::Namespace then super
@@ -87,10 +91,10 @@ module Dry
 
         Dir["#{root}/#{dir}/**/*.rb"].each do |path|
           component_path = path.to_s.gsub("#{dir_root}/", '').gsub('.rb', '')
-          Component.Loader(component_path).tap do |component|
+          Loader(component_path).tap do |component|
             next if key?(component.identifier)
 
-            Kernel.require component.path
+            Kernel.require path
 
             if block_given?
               register(component.identifier, yield(component.constant))
@@ -136,7 +140,7 @@ module Dry
       end
 
       def self.load_component(key)
-        component = Component.Loader(key)
+        component = Loader(key)
         src_key = component.namespaces[0]
 
         if imports.key?(src_key)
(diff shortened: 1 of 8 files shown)


solnic:
Yeah I actually planned to add this feature to be able to use shorter container identifiers. I'll look into this PR once I'm done with other dry-* stuff, so later this month. Thanks for that :)

timriley:
Something like this would be great :+1:

lib/dry/component/container.rb
```diff
@@ -34,6 +34,10 @@ def self.configure(env = config.env, &block)
         self
       end
 
+      def self.Loader(key)
```
davidpelaez: I don't think I follow what this is for here :( Would you mind clarifying the intention? I'm not the one merging this, just reading out of curiosity to learn what you did :)

sagmor: It's a quick hack to be able to change the `Loader` function on a [subclass](https://github.com/sagmor/dry-component/blob/namespace/lib/dry/component/namespace.rb#L35).
I did it like this to touch as little as possible the current container and get the proof of concept working but definitely should be changed.

timriley:
Thanks for your work on this, @sagmor. We've just merged an alternative approach to providing this feature, via #20. Feel free to check it out and let us know what you think!

sagmor:
@timriley Awesome! I'll check it out

timriley:
@sagmor Great :) BTW, right after I shipped this I realised there's probably a better way, which is what I described in https://github.com/dry-rb/dry-component/issues/23

Issue #7, opened 2016-03-11
Root has to be a pathname not a String as indicated in the example


davidpelaez:
I get `dry-component-0.0.1/lib/dry/component/config.rb:9:in `load': undefined method `join' for "app":String (NoMethodError)` when I try the example code:

```ruby
class Application < Dry::Component::Container
  configure do |config|
    config.root = 'app'

    # we set 'lib' relative to `root` as a path which contains class definitions
    # that can be auto-registered
    config.auto_register = 'lib'
  end

  # this alters $LOAD_PATH hence the `!`
  load_paths!('lib')
end
```

I made the `app` string a pathname `config.root = Pathname.new('app')` and it works. Clearly it expect a pathname or assumes it's being instantiated elsewhere. Please let me know if I'm missing something or it's a real bug.

solnic:
Sorry, it's a mistake in the readme but I'd be OK with making it work with strings, too.

davidpelaez:
I'll be nice to have strings. For the sake of exploring and making the best of what you have already released I just sent a small PR for the readme with what I found to be missing in my own experimentation.

PR #8, opened 2016-03-11
Fixes missing readme info about the path

davidpelaez:
After working with the code I found that I needed to pass a Pathname _and_ make it relative to the working directory with `./`. It's a minor thing but hopefully it'll help newcomers experimenting with this.

This is related to #7


solnic:
Thanks!

PR #10, opened 2016-03-21
Allow for config to be erb processed

hmadison:
In some deployment environment setups ([12factor](http://12factor.net/)) the application is expected to read its configuration information from the environment variables present on the system. By erb processing the config file in dry-component, it allows you to read the variables into application config (eg `DATABASE_URL`), and cleanly store constants (eg `max_upload_size`) in source control.

### DIFF
--- a/lib/dry/component/config.rb
+++ b/lib/dry/component/config.rb
@@ -1,4 +1,5 @@
 require 'yaml'
+require 'erb'
 
 module Dry
   module Component
@@ -8,7 +9,8 @@ module Dry
 
         return {} unless File.exist?(path)
 
-        yaml = YAML.load_file(path)
+        data = ERB.new(File.read(path)).result(binding)
+        yaml = YAML.load(data)
 
         Class.new do
           extend Dry::Configurable
--- a/spec/fixtures/test/config/application.yml
+++ b/spec/fixtures/test/config/application.yml
@@ -1,2 +1,3 @@
 test:
   foo: 'bar'
+  env: <%= ENV['USER'] %>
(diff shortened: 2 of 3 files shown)


timriley:
I'm not entirely sure this is the best idea, since `Component::Config` is designed to fetch config from matching ENV vars before falling back to the yml file – see https://github.com/dry-rb/dry-component/blob/master/lib/dry/component/config.rb#L17. Using ERB processing in the values from yml file means that the ENV- and yml-provided data will no longer match.

In your example, you mentioned 2 use cases:

1. Reading ENV vars
2. Cleanly storing constants

As far as I can tell, (1) is already taken care of by the existing `Component::Config` behaviour (the source code I linked to above).

Would you mind taking a bit more time to explain how you see (2) working?

hmadison:
I apologize for not making it clearer but one and two are two benefits produced by case one. Seeing the code you referenced now, it does accomplish a similar goal. I do feel though putting a reference to the environment variable in the yml is less confusing versus merging it in after the yml is loaded.

timriley:
I'll close this one now since we've moved config over into dry-web and made it a lot more flexible, including the ability for you to provide your own entirely custom settings loader (see [this](https://github.com/dry-rb/dry-web/blob/cdd575c37de0b64b1b38a173559b509f70c8c581/lib/dry/web/umbrella.rb) and [this](https://github.com/dry-rb/dry-web/blob/cdd575c37de0b64b1b38a173559b509f70c8c581/lib/dry/web/settings.rb) for more).

PR #11, opened 2016-05-08
Support all dry-auto_inject injector types

timriley:
Support all injector types offered by dry-auto_inject (args, hash, kwargs) while still auto-loading component files as required.

This changes `Container.import_module` to `Container.Inject`, which now returns an `Injector` object that offers API to inject with each of the strategies, e.g.

```ruby
class MyContainer < Dry::Component::Container
end

class MyClass
  # using plain args
  include MyContainer.Inject["foo.bar"]

  # using hash
  include MyContainer.Inject.hash["foo.bar"]

  # using kwargs
  include MyContainer.Inject.kwargs["foo.bar"]
end
```

The `Injector` also supports the new dry-auto_inject dependency aliasing API (e.g. `Inect[local_name: "container.name"]`).

This is a work in progress. I'm putting this PR up now for your feedback!

One change I would like to make is to be able to access the Injector via `::` syntax, e.g. `include MyContainer::Inject[*deps]`, but I'm not quite sure how to do that (or if it is even possible).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,6 +8,7 @@ rvm:
   - 2.0
   - 2.1
   - 2.2
+  - 2.3.1
   - rbx
   - jruby-9000
   - ruby-head
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,7 +8,7 @@ rvm:
   - 2.0
   - 2.1
   - 2.2
-  - rbx-2
+  - rbx
   - jruby-9000
   - ruby-head
   - jruby-head
--- a/Gemfile
+++ b/Gemfile
@@ -2,7 +2,7 @@ source 'https://rubygems.org'
 
 gemspec
 
-gem 'dry-container', github: 'dry-rb/dry-container', branch: 'hook-methods-via-modules'
+gem 'dry-container', github: 'dry-rb/dry-container'
 gem 'dry-auto_inject', github: 'dry-rb/dry-auto_inject'
 
 gem 'codeclimate-test-reporter', platforms: :rbx
--- a/Gemfile
+++ b/Gemfile
@@ -2,6 +2,8 @@ source 'https://rubygems.org'
 
 gemspec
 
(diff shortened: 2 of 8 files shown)


flash-gordon:
@timriley great! I think you can set `Inject` constant in the `inherited` hook. It's gonna be OK because inheritance is the only way for using `Container`.

timriley:
@flash-gordon Ah, yes, great idea. I've just pushed this now.

I think it feels a lot nicer if you're using the component's injector directly, e.g.

```
class MyClass
  # plain injector:
  include MyContainer::Inject["foo.bar"]
  
  # or other types:
  include MyContainer::Inject.kwags["foo.bar"]
end
```

In dry-web/rodakase apps, and even in the test suite for dry-component, this injector is aliased to a more directly accessible constant (e.g. `MyApp::Import = MyApp::Container::Inject`), but I still think it's nice to have things arranged so that direct use is also nice 😄

timriley:
Some other notes:

* Setting the `Inject` constant means we can ditch the memoization in `Dry::Component::Container.injector`, which I think is a nice improvement, since `Inject` is implicitly memorised, given it's a single instance, and it's intended to be the main interface for the user to inject dependencies from the component's container.
* I've used the "inject" name instead of "import" because `Dry::Component::Container` already has methods named with "import", and they are for importing _other_ containers into this one. I wanted to avoid any confusion. Users of the dry-component container can always create their own aliases that make more sense (e.g. like dry-web apps with their own `Import` constants).

This feels to me like a decent initial API design. Unless there's any feedback requiring changes, I'll move onto adding some test cases to cover the injector working with different injection types (hash, kwargs, etc.)

timriley:
This is ready to go. Tests in place and API actually aligns with [my notes here](https://github.com/dry-rb/dry-component/pull/11#issuecomment-217756843). Will merge as soon as rbx passes on Travis.

solnic:
😍 😍 😍 😍 😍 😍 😍 😍

PR #13, opened 2016-06-06
Allow for a configurable loader

AMHOL:


timriley:
This looks good to me, @AMHOL :)

flash-gordon:
Really nice. Things look simpler now. I have an idea about using configurable loader with namespace like this
```ruby
module MySpace
  class Application < Dry::Component::Container
    configure do |config|
      config.loader = Dry::Component::Loader.namespace(:my_space)
      # or
      config.loader = Dry::Component::Loader[MySpace]
    end
  end
end
```
We can use inheritance for that or sutin

timriley:
@flash-gordon A namespace-aware loader sounds really great. I'd love to make that part of the default dry-web setup.

PR #17, opened 2016-06-16
Remove config and env

### LINKED ISSUE #14: Move config to dry-web
Right now, dry-component’s config is a little under-baked, suffering from a few different things:

- The actual `.yml` files _must_ exist on the filesystem before config keys can exist, even if config values are available in `ENV` since the `.yml` files act as the canonical list of the available keys. This causes problems when making 12-factor-style deployments, since we have to keep this config file checked in, which means it makes it awkward to use secret config values during development.
- Config files are looked up relative to the component’s root, and in a typical app where multiple components are installed, this means config files are spread around the filesystem, which is unexpected
- Config files being named after the component’s name also makes them harder to find
- And, biggest of all, in a typical dry-web app, _only the config for the top-level component is really used_. This means we have a ton of sub-components scattered around the place with never-used config systems.

I do recognise that the config system in its current state wasn’t intended to be the final solution :) However, I think because of the issues above it may be a point of confusion for people coming to build a dry-web app.

I want to propose an interim step while we develop a more complete config system:

**Remove config from dry-component and move it to dry-web.**

This would mean that a dry-web app would have a single point of configuration, which would be accessible from any dry-component used within it. We’d make this available on `Dry::Web::Container`, and I think it’d be smart to make the config class configurable just like @AMHOL’s done with the dry-component loader. That way we could provide a very simple default config class, and allow people to replace it with their own (e.g. in our apps at Icelab we’re using a config class where the keys are explicitly specified and validated for appropriate values).

With this in place, I think a dry-web app will be easier to understand, and then we can more gradually work towards an per-component config approach. There are a number of things that I think would be helpful there, like a component specifying what config keys it requires, and then those keys being plucked out of a top-level config object that is passed to the component when it is booted. But these are the things that’d take a little experimentation to get right, which is why I think it’d be good to settle on a reasonable interim arrangement :)

What do you think, @solnic, @AMHOL? This is one of the things I'd love to straighten out a little before RDRC.

timriley:
As detailed in #14, the current config object in dry-component has a number of shortcomings, and isn't really used outside of one specific (a single top-level component) place in existing dry-web apps.

This PR:

- Removes config loading from dry-component's `Container`
- Also removes the `env` setting

Both of these will be handled by dry-web now (and in a single place – the dry-web `Umbrella`), which will make dry-web apps easier to understand and deploy. Then, later, we can look at a more appropriate way to hand config around to particular dry-components.

Resolves #14


timriley:
Since we're in agreement about this in #14, I'm going to merge this now.

PR #19, opened 2016-07-11
Support all auto inject strategies

timriley:
Allow `Dry::AutoInject` options to be passed to `Dry::Component::Container#injector`. This allows the underlying auto_inject object to be configured as the user requires, while still retaining dry-component's desired behaviour of lazily loading dependencies.

The immediate benefit of this is that a user can now set up an injector with their own custom strategies container. However, it does mean that any options added to dry-auto_inject in the future will automatically be available for dry-component's injector as well.

This requires a breaking behavioural change, though. The `Inject` constant that we were setting up automatically on `Dry::Component::Container` subclasses can't be set up in the same way anymore, since there's no way we can infer by that time if the user wants any particular options to be passed. The recommended behaviour, instead, would be to have the user explicitly define their own constant to store their own preferred injector (e.g. `MyApp::Inject = MyApp::Container.injector(my_options)`). I think this is a reasonable requirement of users, though, and it was actually what we required of them until the last 0.3.0 release of dry-component, anyway.

Sound OK?

### DIFF
--- a/lib/dry/component/injector.rb
+++ b/lib/dry/component/injector.rb
@@ -2,7 +2,7 @@ require "dry-auto_inject"
 
 module Dry
   module Component
-    class Injector
+    class Injector < BasicObject
       # @api private
       attr_reader :container
 
@@ -10,14 +10,13 @@ module Dry
       attr_reader :injector
 
       # @api private
-      def initialize(container, strategy: nil, strategies_cache: nil)
+      def initialize(container, strategy: nil)
         @container = container
-        @strategies = strategies_cache
         @injector =
           if strategy
-            Dry::AutoInject(container).send(strategy)
+            ::Dry::AutoInject(container).send(strategy)
           else
-            Dry::AutoInject(container)
+            ::Dry::AutoInject(container)
           end
       end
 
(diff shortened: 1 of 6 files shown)


lib/dry/component/injector.rb
```diff
@@ -27,37 +25,24 @@ def [](*deps)
...
-        strategies[:hash]
+      def method_missing(name, *args, &block)
+        ::Dry::Component::Injector.new(container, options: options, strategy: name)
       end
 
-      # @api public
-      def kwargs
-        strategies[:kwargs]
+      def respond_to_missing?(name, _include_private = false)
+        injector.respond_to?(name)
```
AMHOL: missing `|| super` here

lib/dry/component/injector.rb
```diff
@@ -2,23 +2,21 @@
...
         @container = container
-        @strategies = strategies_cache
-        @injector =
-          if strategy
-            Dry::AutoInject(container).send(strategy)
-          else
-            Dry::AutoInject(container)
-          end
+        @options = options
+        @injector = ::Dry::AutoInject(container, options).send(strategy)
```
AMHOL: Can this use `#public_send` ?

lib/dry/component/injector.rb
timriley: I've realised that this should actually just be `#respond_to?` inside a `BasicObject` subclass (since it doesn't actually offer `#respond_to?` by default). Will push up this adjustment shortly.

lib/dry/component/injector.rb
timriley: It can be! Thanks for the reminder :)

timriley:
I'll merge this in now since all the code tweaking is done and there was no issue raised with the API adjustment :)

PR #20, opened 2016-07-12
Add configurable default namespace to containers

timriley:
When a user sets a default namespace on a container, that namespace will not be used in any container registrations for components from within that namespace. This makes it easier to reference them later.

For example, in a container with a namespace of “foo.bar”, a component at “foo/bar/baz.rb” will be accessible on the identifier “baz” only (with the default namespace removed for convenience).

There are a few design adjustments to make this possible:

- The container's configured `loader` is instantiated with the container object being passed to it. This allows the loader to inspect the container for config and various other settings.
- The `loader` object no longer represents a single loaded component file, but instead should offer `#load(file)`, which _then_ returns an object to represent that loaded component file. This is what makes the previous change possible, making the loader a "service" which can inspect the container's config etc. as required.

Along with these structural changes, our default loader now checks for a `default_namespace` setting and will remove that from the front of any container identifiers that match. This gives us the ability to configure a container with `"my_app.foo"` as a namespace, and then access the component file `my_app/foo/bar.rb` as `Container["bar"]`. Yay!

It's probably worth comparing this to https://github.com/dry-rb/dry-component/pull/6 (👋 @sagmor) to see if there's anything from there we'd like to pull across or incorporate (or if we prefer that other implementation altogether).

What do you reckon?

### DIFF
--- a/lib/dry/component/loader.rb
+++ b/lib/dry/component/loader.rb
@@ -11,8 +11,8 @@ module Dry
           @loader = loader
 
           @identifier = input.to_s.gsub(loader.path_separator, loader.namespace_separator)
-          if loader.namespace
-            re = /^#{Regexp.escape(loader.namespace)}#{Regexp.escape(loader.namespace_separator)}/
+          if loader.default_namespace
+            re = /^#{Regexp.escape(loader.default_namespace)}#{Regexp.escape(loader.namespace_separator)}/
             @identifier = @identifier.gsub(re, '')
           end
 
@@ -41,13 +41,13 @@ module Dry
 
       PATH_SEPARATOR = '/'.freeze
 
-      attr_reader :namespace
+      attr_reader :default_namespace
       attr_reader :namespace_separator
       attr_reader :path_separator
 
-      def initialize(container)
-        @namespace = container.config.namespace
-        @namespace_separator = container.config.namespace_separator
+      def initialize(config)
+        @default_namespace = config.default_namespace
+        @namespace_separator = config.namespace_separator
         @path_separator = PATH_SEPARATOR
       end
 
--- a/lib/dry/component/loader.rb
+++ b/lib/dry/component/loader.rb
@@ -3,33 +3,56 @@ require 'inflecto'
 module Dry
   module Component
     class Loader
-      IDENTIFIER_SEPARATOR = '.'.freeze
+      class Component
+        attr_reader :loader
+        attr_reader :identifier, :path, :file
+
+        def initialize(loader, input)
+          @loader = loader
+
+          @identifier = input.to_s.gsub(loader.path_separator, loader.namespace_separator)
+          if loader.namespace
+            re = /^#{Regexp.escape(loader.namespace)}#{Regexp.escape(loader.namespace_separator)}/
+            @identifier = @identifier.gsub(re, '')
+          end
+
+          @path = input.to_s.gsub(loader.namespace_separator, loader.path_separator)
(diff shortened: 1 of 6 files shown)


lib/dry/component/loader.rb
```diff
@@ -3,33 +3,56 @@
...
+
+      attr_reader :namespace
+      attr_reader :namespace_separator
+      attr_reader :path_separator
+
+      def initialize(container)
+        @namespace = container.config.namespace
+        @namespace_separator = container.config.namespace_separator
+        @path_separator = PATH_SEPARATOR
+      end
```
solnic: It doesn't seem to be a good idea to couple `Loader` with an entire container like that. Why not just pass in the config?

spec/unit/container/auto_register_spec.rb
```diff
@@ -18,15 +18,38 @@ class Test::Container < Dry::Component::Container
...
-          super.gsub('.', '-')
+        class Component < Dry::Component::Loader::Component
+          def identifier
+            super + ".yay"
+          end
+
+          def instance(*args)
+            constant.respond_to?(:call) ? constant : constant.new(*args)
+          end
         end
```
solnic: this is *awesome*, feels like we could leverage that for our monitoring features

lib/dry/component/loader.rb
timriley: That'd work too. I suppose we could actually have a loader-specific config namespace that we pass so only appropriate config gets sent along. Seem better?

solnic: Yes that would be perfect

flash-gordon:
awesome!

lib/dry/component/loader.rb
```diff
@@ -3,33 +3,56 @@
...
 
-      def initialize(input)
-        @identifier = input.to_s.gsub(PATH_SEPARATOR, IDENTIFIER_SEPARATOR)
-        @path = input.to_s.gsub(IDENTIFIER_SEPARATOR, PATH_SEPARATOR)
-        @file = "#{path}.rb"
-      end
+          @identifier = input.to_s.gsub(loader.path_separator, loader.namespace_separator)
+          if loader.default_namespace
+            re = /^#{Regexp.escape(loader.default_namespace)}#{Regexp.escape(loader.namespace_separator)}/
+            @identifier = @identifier.gsub(re, '')
```
AMHOL: Could this not just use sub as the regex is anchored?

timriley: Yes, totally. I'll fix!

lib/dry/component/loader.rb
```diff
@@ -3,33 +3,56 @@
...
+          end
 
-      def namespaces
-        identifier.split(IDENTIFIER_SEPARATOR).map(&:to_sym)
-      end
+          @path = input.to_s.gsub(loader.namespace_separator, loader.path_separator)
```
AMHOL: http://ruby-doc.org is down but I believe you could use `String#trtr` here?

timriley: > `tr(from_str, to_str) => new_str`
> Returns a copy of str with the characters in from_str replaced by the corresponding characters in to_str. If to_str > is shorter than from_str, it is padded with its last character in order to maintain the correspondence.

I wonder if this could get us into trouble if the namespace separator was more than one character long? Perhaps gsub is still the safest bet here?

AMHOL: Yeah you're right, best not to make that assumption

Issue #23, opened 2016-07-26
Consider alternative default namespace approaches


timriley:
In implementing this feature and having to turn around the quick-fire fix in #22, I realised there might be some better ways to do this.

I was thinking that it might be better for the default namespace handling to be done on the _resolution_ side rather than the _registration_ side. dry-component could have its own resolver that is default namespace aware, so it would know to convert "foo" to "main.foo" if "main" was the default namespace and there was a matching registration.

For a convenience, this approach seems a lot cleaner and more efficient than getting involved in the actual file loading process. It also means all the container registrations remain "canonical" and fully matching their corresponding file paths, etc.

I was also thinking that it might be good if we can make it _explicit_ when a user intends to access a container registration as in the shorthand format (i.e. with the default namespace removed):

```ruby
# the leading `.` here implies the shorthand format
include Admin::Import[".users.repository"]
```

In this case, if `Admin::Import` is configured with "admin" as the default namespace, then it would know 100% that ".users.repository" should be converted to "admin.users.repository" and looked up accordingly.

What do you think? This feels like it would remove any possible ambiguity, which'd be a good thing.

AMHOL:
I feel like the leading `.` should do the opposite and indicate that it should not be prefixed - same as the scope resolution operator (`::`).

timriley:
@AMHOL interesting perspective! I had not thought of that.

I guess I see the leading dot acting kind of like how you chain methods across lines in Ruby:

```ruby
  foo(something)
    .bar
```

The ".bar" here, with it's leading dot, is _clearly_ "attached" to the item that preceded it. This is what I was going for with the shorthand container identifier syntax. WDYT?

AMHOL:
I still think with the `.` being the namespace separator, it would be more intuitive to have it act as the scope-resolution operator. To imply that it is qualified with something would just confuse me personally.

timriley:
I just realised I had `include Admin::Container[".users.repository"]` in my top description when I meant it to be `include Admin::Import[".users.repository"]`, i.e. the injector for a dry-component's container. I've corrected it now.

I have a feeling this correction wouldn't change your thoughts on this though, and that's fair enough, let's try and avoid anything that's potentially confusing.

So perhaps the best approach is to offer the same lookup as we offer now, with `"users.repository"` returning the object registered as `"admin.users.repository"` when `admin` is the component's default namespace.

[...]

PR #27, opened 2016-08-08
Refactor Container/Component/Loader

solnic:
This improves separation of concerns and allows lazy-loading bootable deps.

## Summary

- [x] Move component-related logic from `System::Container` to `System::Component`
- [x] Refactor `System::Loader` into a simple class that is only responsible for object instantiation and is injected into components
- [x] Extend `System::Component` so that it can be namespace-aware
- [x] Make it possible that import module can trigger booting of a dependency

### DIFF
--- a/lib/dry/system/component.rb
+++ b/lib/dry/system/component.rb
@@ -29,6 +29,24 @@ module Dry
         freeze
       end
 
+      def bootable?(path)
+        boot_file(path).exist?
+      end
+
+      def boot_file(path)
+        path.join("#{root_key}.rb")
+      end
+
+      def file_exists?(paths)
+        paths.any? { |path| path.join(file).exist? }
+      end
+
+      def prepend(name)
+        self.class.new(
+          [name, identifier].join(separator), options.merge(loader: loader.class)
+        )
+      end
+
       def namespaced(namespace)
         self.class.new(
           path, options.merge(loader: loader.class, namespace: namespace)
@@ -41,11 +59,6 @@ module Dry
 
       def namespace
         options[:namespace]
-      end
-
-      def dependency?(name)
-        *deps, _ = namespaces
-        (deps & name.split(separator).map(&:to_sym)).size > 0
       end
 
       def root_key
--- a/lib/dry/system/component.rb
+++ b/lib/dry/system/component.rb
@@ -7,19 +7,14 @@ module Dry
       include Dry::Equalizer(:identifier, :path)
 
       PATH_SEPARATOR = '/'.freeze
+      WORD_REGEX = /\w+/.freeze
 
       attr_reader :identifier, :path, :file, :options, :loader
 
       def self.new(name, options)
-        ns, sep = options.values_at(:namespace, :separator)
+        ns, sep = options.values_at(:namespace, :separator).map(&:to_s)
 
-        identifier =
-          if ns
-            name.to_s.sub(%r[^#{ns}#{sep}], '')
-          else
-            name.to_s.gsub(PATH_SEPARATOR, sep)
-          end
-
+        identifier = name.to_s.scan(WORD_REGEX).reject { |s| ns == s }.join(sep)
         path = name.to_s.gsub(sep, PATH_SEPARATOR)
         loader = options.fetch(:loader, Loader).new(path)
 
@@ -32,6 +27,12 @@ module Dry
         @file = "#{path}.rb".freeze
         @loader = options.fetch(:loader)
         freeze
+      end
+
+      def namespaced(namespace)
+        self.class.new(
+          path, options.merge(loader: loader.class, namespace: namespace)
+        )
       end
 
       def separator
@@ -51,12 +52,14 @@ module Dry
         namespaces.first
       end
 
-      def namespaces
-        identifier.split(separator).map(&:to_sym)
-      end
-
       def instance(*args)
         loader.call(*args)
+      end
+
+      private
+
+      def namespaces
+        identifier.split(separator).map(&:to_sym)
       end
     end
   end
--- a/lib/dry/system/component.rb
+++ b/lib/dry/system/component.rb
@@ -14,6 +14,8 @@ module Dry
       def self.new(name, options)
         ns, sep = options.values_at(:namespace, :separator).map(&:to_s)
 
+        raise InvalidNamespaceError, ns if ns.include?(sep)
+
         identifier = name.to_s.scan(WORD_REGEX).reject { |s| ns == s }.join(sep)
         path = name.to_s.gsub(sep, PATH_SEPARATOR)
         loader = options.fetch(:loader, Loader).new(path)
--- a/lib/dry/system/component.rb
+++ b/lib/dry/system/component.rb
@@ -1,3 +1,5 @@
+require 'concurrent/map'
+
 require 'dry-equalizer'
 require 'dry/system/loader'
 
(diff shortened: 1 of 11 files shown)


spec/unit/component_spec.rb
```diff
@@ -0,0 +1,42 @@
...
+    describe '#identifier' do
+      it 'returns qualified identifier' do
+        expect(component.identifier).to eql('foo')
+      end
+    end
+
+    describe '#namespaces' do
+      it 'returns namespace array' do
+        expect(component.namespaces).to eql(%i[foo])
+      end
```
timriley: Wouldn't we want namespaces to be `[]` for a component with a single name not nested under anything?

solnic: Yes we do, I noticed that as well. It's just an unfortunate naming. I'll be refining this API. Stay tuned :)

timriley: Such suspense! 😉

lib/dry/system/component.rb
```diff
@@ -1,49 +1,78 @@
...
-        if loader.default_namespace
-          re = /^#{Regexp.escape(loader.default_namespace)}#{Regexp.escape(loader.namespace_separator)}/
-          @identifier = @identifier.sub(re, '')
-        end
+      def self.new(name, options)
+        ns, sep = options.values_at(:namespace, :separator).map(&:to_s)
 
-        @path = input.to_s.gsub(loader.namespace_separator, loader.path_separator)
-        @file = "#{path}.rb"
+        identifier = name.to_s.scan(WORD_REGEX).reject { |s| ns == s }.join(sep)
```
AMHOL: What's this for? Doesn't seem right to me

lib/dry/system/container.rb
```diff
@@ -72,9 +77,9 @@ def self.auto_register!(dir, &_block)
         dir_root = root.join(dir.to_s.split('/')[0])
 
         Dir["#{root}/#{dir}/**/*.rb"].each do |path|
-          component_path = path.to_s.gsub("#{dir_root}/", '').gsub('.rb', '')
+          path = path.to_s.gsub("#{dir_root}/", '').gsub('.rb', '')
```
AMHOL: Could these be anchored either side and use `#sub`?

lib/dry/system/component.rb
solnic: It's not as robust as it could be, but in general this creates a component identifier just like it did before. ie it will do `foo/bar/` => `foo.bar` and if `:foo` is passed as the ns, then it'll end up with `bar`.

lib/dry/system/container.rb
solnic: WDYM? :)

AMHOL: Like:

```ruby
path.to_s.sub(/^#{dir_root}\//, '').sub(/\.rb$/, '')
```

lib/dry/system/component.rb
AMHOL: Ahh, OK, what about if you wanted to pass a namespace like `foo.bar` and turn `foo.bar.baz` into `baz/`, don't think that would work? Wouldn't it also convert `foo.bar.foo` into `bar/` with a namespace of `foo`?

solnic: > what about if you wanted to pass a namespace like foo.bar and turn `foo.bar.baz` into `baz/`

We shouldn't allow that, the ns should not contain the separator

> Wouldn't it also convert `foo.bar.foo` into `bar/` with a namespace of `foo`?

It would turn it into `bar` actually, which would break.

...that's what I meant by "its not as robust as it could be" :) I'll add proper validation later though

AMHOL: COol :+1:

solnic:
@AMHOL @timriley k so this has gotten too big so I'm stopping already...please take the final look and let's get this merged in :)

AMHOL:
:+1:

lib/dry/system/component.rb
```diff
@@ -1,49 +1,95 @@
...
+
+      PATH_SEPARATOR = '/'.freeze
+      WORD_REGEX = /\w+/.freeze
+
+      attr_reader :identifier, :path, :file, :options, :loader
+
+      def self.new(*args)
+        cache.fetch_or_store(args.hash) do
+          name, options = args
+          ns, sep = options.values_at(:namespace, :separator).map(&:to_s)
```
timriley: Do we ever initialize components without an options hash? e.g. `Dry::System::Component.new("foo")`? If so, then this line crashes with a `NoMethodError: undefined method `values_at' for nil:NilClass`.

timriley: I'm guessing not, since this is an "internal" class. I was just doing some playing around on the console and discovering the API as I went, which is why I hit these errors (I also tried ` Dry::System::Component.new("foo", {})`, which got me an `InvalidNamespaceError`, since the separator in that case becomes `nil.to_s`, a blank string, which conflicts with a `nil.to_s` namespace).

timriley: If we _did_ want to make component initialization more robust, I'd recommend a default `"."` value for the separator, and allowing it to be created without an options hash.

lib/dry/system/component.rb
```diff
@@ -1,49 +1,95 @@
-require 'inflecto'
+require 'concurrent/map'
+
+require 'dry-equalizer'
+require 'dry/system/loader'
```
timriley: Do we need a require for "dry/system/errors" here? When I required this file individually, I was setting errors loading our errors 😆 e.g. `NameError: uninitialized constant Dry::System::Component::InvalidComponentError`

timriley:
Just getting my head around things...

I wonder if we should be raising an exception for a component like this?

```
Dry::System::Component.new('foo', namespace: 'foo', separator: '.')
```

This ends up having an identifier of `""`, which I presume isn't something we want?

timriley:
With the renames, I wonder if we should make `'system'` the default `:core_dir` setting?

timriley:
The default namespace handling here is nice 👏 Totally handles what I was thinking about in #23.

lib/dry/system/loader.rb
```diff
@@ -10,8 +10,8 @@ def initialize(path)
       end
 
       def call(*args)
-        if constant.respond_to?(:instance) && !constant.respond_to?(:new)
-          constant.instance(*args) # a singleton
+        if singleton?(constant)
```
timriley: Couldn't help but make a tweak 😉 /cc @solnic

spec/integration/import_spec.rb
```diff
@@ -0,0 +1,63 @@
...
+      expect(user_repo.repo).to be_instance_of(Db::Repo)
+    end
+
+    it 'loads an external dep during finalization' do
+      system.finalize!
+      expect(user_repo.repo).to be_instance_of(Db::Repo)
+    end
+  end
+
+  context 'when top-level container provides the dependency' do
```
timriley: Is this example description mixed up with the one below? Seeing that the user repo here is importing `core.db.repo`, I thought the description would be 'when top-level container requires the dependency _from the imported container_', since `core` denotes the imported container?

timriley:
This is a great set of changes! Apart from the few specific questions/comments, it all looks good to me 👍 

I do have one related thought, though. I feel slightly uncomfortable with this behaviour (pulling this code from `import_spec.rb`):

```ruby
class Umbrella < Dry::System::Container
  configure do |config|
    config.name = :core
  end
end

  # ...
end

App.import(Umbrella)
```

It feels like there's too much happening implicitly when we import `Umbrella` and have its name inside `App` automatically set for us based on Umbrella's configured `name` setting.

IMO it might be better if we required the user to explicitly set a name on import, e.g.

```ruby
App.import(shared: Umbrella)
```

Especially since the user developing `App` will have to decide on the namespace they want to use for the imported container ("shared" in this example here). If they let the `Umbrella` app choose its own name, then the whole off the `App` container could be disrupted by a change made _outside of it_ (i.e. changing the name setting inside `Umbrella`), which doesn't feel like the most ideal separation of responsibilities/configuration.

I know this `import(foo: Bar)` behaviour is already supported. I'm wondering whether we should promote that as the preferred path (and potentially deprecate/remove the implicit setting of the name).

lib/dry/system/loader.rb
solnic: It's fine :) I left the original code as I'm actually planning to figure out a way to get rid of the conditional entirely, but not as part of this PR.

solnic:
> With the renames, I wonder if we should make 'system' the default :core_dir setting?

Yes I wanted to do it but first I wanted to ask you if you're OK with that, I guess I know now :)

solnic:
> IMO it might be better if we required the user to explicitly set a name on import

Yep. I agree. Do you think it's worth the effort to deprecate it? I'd just change it and add a note in the CHANGELOG and call it a day.

timriley:
> Yep. I agree. Do you think it's worth the effort to deprecate it? I'd just change it and add a note in the CHANGELOG and call it a day.

Cool, let's do it then :) I think a CHANGELOG mention is fine, no need for a formal deprecation cycle.

solnic:
@timriley let's make further adjustments in master then. I'll merge this in and test it against Berg first.

PR #28, opened 2016-08-16
Standalone example needs 'bundler/setup' to run

rickenharp:
Running the standalone example results in an error:
`cannot load such file -- dry/system/container (LoadError)`

This PR adds the necessary require statement to get the bundler environment up and running.


solnic:
oh thank you! I'm so used to running stuff with `bundle exec` that I just forgot about this :)

Issue #29, opened 2016-08-22
Invoking undefined methods does not cause exceptions.


aradunovic:
Today, I stumbled upon some weird behaviour while trying to use dry-system. Here's the code in question:

```ruby
# system/container
require "dry/system/container"

class Container < Dry::System::Container
  configure do |config|
    config.auto_register = "lib"

    load_paths!("lib")
  end
end

# system/boot/db.rb
Container.finalize(:db) do |container|
 
  _unexisting_method # THIS SHOULD FAIL

  use :logger
  require "sequel"

  Sequel.database_timezone = :utc
  Sequel.application_timezone = :local
  db = Sequel.connect(ENV["DB_URL"])
  db.loggers << logger

  container.register(:db, db)
end

# system/boot.rb
require_relative "container"
Container.finalize!
```

Result:

```ruby
require_relative "system/boot"

# SUCCESS
Container["db"].tables   #=> schema_info
```

Even though I tried to invoke the undefined method, no exceptions were raised.

solnic:
That's a simple fix, fortunately, we just need to add `else super` to [`method_missing`](https://github.com/dry-rb/dry-system/blob/master/lib/dry/system/lifecycle.rb#L95-L101) :)

aradunovic:
Hey, I'll be happy to contribute, if you're too busy. The fix is brutally simple. 
Now, the only thing I'm not so sure about is how should we test this?

solnic:
That would be very helpful. You can simply add an example to [this spec](https://github.com/dry-rb/dry-system/blob/master/spec/unit/container/finalize_spec.rb) - something like this should be enough:

``` ruby
expect {
  system.finalize(:db) do
    oops('hah')
  end
  system.booter.boot!(:db)
}.to raise_error(NoMethodError, /oops/)

expect {
  system.finalize(:db) do
    oops
  end
  system.booter.boot!(:db)
}.to raise_error(NameError, /oops/)
```

solnic:
Fixed via #30 and released in 0.5.1

PR #30, opened 2016-08-23
Fix silent failures on undefined methods or variables (#29)

aradunovic:
@solnic I couldn't improve/simplify the spec examples you [suggested](https://github.com/dry-rb/dry-system/issues/29#issuecomment-241490760). More specifically, I intended to extract the finalization and boot code and DRY things up a bit. However, this would require additional helper method(s) and that seemed like an overkill to me.
Also, I hope the example description is good.

BTW, I believe`spec/fixtures/test/log/test.log` should be ignored, right?


solnic:
> BTW, I believespec/fixtures/test/log/test.log should be ignored, right?

Right!

PR #36, opened 2016-09-28
Perform registration even though component already required

### LINKED ISSUE #34: Auto register and already required files
Hi all :v:

We've encountered some unexpected behaviour with auto register.

Let's say I have `Services` folder with 2 files:

```ruby
# 1.rb
class One
end
```

```ruby
# 2.rb
require './1.rb'
class Two
end
```

Both files are inside of some auto register folder. And in this case load order matters, cause if `2.rb` is auto registered (and so required) first, `1.rb` will not be registered at all because [`Kernel.require(component.path)`](https://github.com/dry-rb/dry-system/blob/master/lib/dry/system/container.rb#L454) will return false. Preventing [this code](https://github.com/dry-rb/dry-system/blob/master/lib/dry/system/auto_registrar.rb#L29) from registering component.

So, maybe `Kernel.require(component.path) && yield` should be replaced with:

```ruby
Kernel.require(component.path)
yield
```

Otherwise why it is done with `&&`?

Thanks for help

### LINKED ISSUE #35: [discussion] Find a way to have component registered even though it was previously required
Hi!

I've recently entered an issue of which it seems some changes need to be applied [here][1] in order to get it to work. I mean the possiblity to register component even though it was required somewhere else. Consider inheritance as an example – either base class has to be registered first so the order of registering components _must be_ explicit or an extra `require` statement is needed in the subclass so the parent constant is known. Although `Kernel.require(component.path) && yield` is very clever I suppose it may be inconsistent due to the [guard][2] statement present at the method beggining – in essence, if the component exists, go away. Actually the block shouldn't ever be performed if key associated to given component already exists. However, I may not be familiar with all of the use cases, but simple change `Kernel.require(component.path); yield` vastly changes the whole registration process.

From what I've found, the `yield` concerns only use cases when component has to be registered, e.g. [here][3] and [here][4], however in this case [guard][3] forbids registration if a component already exists so we're covered, or at least that's my way of thinking.

The current container apporach prefers aggregation, whereas in certain cases inheritance might be a solution as well. 

What do you guys think about it ?

ps. I ran `Kernel.require(component.path); yield` in my local environment and didn't get into any troubles. The `dry-system` specs have passed as well.

[1]: https://github.com/dry-rb/dry-system/blob/34de07d1ad5f7d0344b9610877334eca74cd70ff/lib/dry/system/container.rb#L454
[2]: https://github.com/dry-rb/dry-system/blob/34de07d1ad5f7d0344b9610877334eca74cd70ff/lib/dry/system/container.rb#L448
[3]: https://github.com/dry-rb/dry-system/blob/34de07d1ad5f7d0344b9610877334eca74cd70ff/lib/dry/system/container.rb#L481
[4]: https://github.com/dry-rb/dry-system/blob/902c0f3c4f6313502ba0f50e68fec51066b33943/lib/dry/system/auto_registrar.rb#L31

blelump:
Hi ! 

Here's a small change discussed in #35 

Closes #35 
Closes #34

### DIFF
--- a/lib/dry/system/container.rb
+++ b/lib/dry/system/container.rb
@@ -451,7 +451,9 @@ module Dry
             raise FileNotFoundError, component
           end
 
-          Kernel.require(component.path) && yield
+          Kernel.require(component.path)
+
+          yield
         end
 
         # @api private
--- a/spec/unit/container_spec.rb
+++ b/spec/unit/container_spec.rb
@@ -34,13 +34,26 @@ RSpec.describe Dry::System::Container do
     end
 
     describe '.require_component' do
-      it 'requires component file' do
-        component = container.component('test/foo')
-        required = false
-        container.require_component(component) do
-          required = true
+
+      shared_examples_for 'requireable' do
+        it 'requires component file' do
+          component = container.component('test/foo')
+          required = false
+          container.require_component(component) do
+            required = true
+          end
+          expect(required).to be(true)
         end
-        expect(required).to be(true)
+      end
+
+      it_behaves_like 'requireable'
+
+      context 'when already required' do
+        before do
+          Kernel.require('test/foo')
+        end
+
+        it_behaves_like 'requireable'
       end
 
       it 'raises when file does not exist' do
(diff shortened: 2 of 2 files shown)


blelump:
Uhm, for some reason Rubinius build fails, but it seems unrelated.

solnic:
Could you rebase on top of master and add a test for this use case?

blelump:
@solnic , sure, done.

solnic:
Thanks. Are you planning to work on a test too?

blelump:
@solnic , aww right. Sorry I've missed that part – will add it tomorrow.

blelump:
@solnic , added. Not sure on that [part][1] though, because it hits container directly.

[1]: https://github.com/dry-rb/dry-system/pull/36/commits/0b502d9fbbb55e0a19ea26930c24c7ec58af9b56#diff-662a8ca0583951270ed8221c1ef6b5b4R66

vladra:
Also closes #34

solnic:
Merged. Thank you :)

Issue #33, opened 2016-09-14
Manually registered container items aren't lazily loaded


timriley:
We need a convention (FS location and files named based on namespace?) to pick up "manual registration" files and require them if we can't otherwise find a component when trying to lazy load it.

timriley:
So actually, this is an issue that spans a number of things.

One of the issues is how dry-validation allows you to pass default options to your schemas:

```ruby
MyForm = Dry::Validation.Schema do
  configure do
    option :foo, Container["bar.baz"]
  end
end
``` 

With this arrangement, `Container.[]` is used to find the item, rather than the container's injector, and `.[]` doesn't do any kind of lazy component loading at all.

timriley:
In fact, I wonder if we could move the lazy loading behavior out from he injector and into the container resolver itself? i.e., enabling it to work when using `System::Container.[]`

AMHOL:
@timriley I was thinking about this recently and came to the same conclusion, I think a custom resolver would be a good suit for this

timriley:
Cool. Yeah, I reckon we could just transplant whatever is inside the `Injector` class and move it into a resolver.

Then later on we can think about extra logic to support a conventional place for manually registered items, etc.

solnic:
This sounds like a good plan :)

davidpelaez:
Is this why stubbing doesn't work after booting a dependency? I'm not sure. If I stub a manually registered item after booting the dependency that uses it the behavior doesn't change. e.g.

```ruby
      Application.boot :operations # resources.find_transaction is registered here manually as a proc
      Application.stub 'resources.find_transaction', -> *x {
        raise "y" # Never called
      }
```

If I stub and the boot it'll raise as expected. Am I correct in that this is the source of the issue? The best solution in the meantime would be to use auto registration?

solnic:
Uhm, you know you can do `option :foo, -> { App['foo'] }` right? :)

solnic:
> @davidpelaez Is this why stubbing doesn't work after booting a dependency?

That's unrelated, please report an issue if you have trouble stubbing a dep that was registered through booting, just remember that if you finalized a container stubbing won't work at all at the moment as we freeze the container unconditionally (I mentioned in another issue that this can be easily solved by skipping `freeze`-ing when we're in a test mode).

timriley:
[...]

PR #39, opened 2016-10-09
Lazy load components as they are resolved, rather than on injection

timriley:
Over in #33 we discussed how it might be better to load a system's components whenever they are resolved, rather than just when they were injected, since it'll allow the lazy loading behaviour to work in more circumstances.

Here's a first take at it. I initially tried to do it with a custom "resolver" for the `Dry::Container` that `Dry::System::Container` inherits from, but it didn't quite give me what I need, since Dry::Container's resolver class doesn't actually have access to the container object itself, just its internal storage hash, and I needed access to the system object so I could call `#load_component`.

Anyway, I realised that all I _really_ needed was a custom `#resolve` on `Dry::System::Container` with a single extra line added, to call `#load_component`.

What do you think, @solnic, @AMHOL? This seem fair to incorporate?

The only thing I don't like is the begin/rescue. I felt like it would be better for a failure to load the component to be manifested as an actual `Dry::Container::Error` "nothing registered" error rather than a dry-system `ComponentLoadError`, since the auto-loading is really meant to be a side-effect of calling `#resolve`, not the actual outcome. Can you think of a way to do this more nicely?


timriley:
One other benefit of this approach: we can totally rip out our custom `Injector` class, which is nice!

blelump:
Hi @timriley ,

I've landed here, because it seems I'm having similar problems, so I've pulled this PR to see how it works. It indeed lazy loads dependencies, however what if a container imports another container (actually this is what I'm doing)? Without finalization, import doesn't provide any dependency and yet, it doesn't know that 'parent' container should load that particular dependency, so in the end it cannot find suitable one and blows up.

AMHOL:
I'm not sure about the error suppression TBH, if we are implicitly loading I think that I'd rather see a loading error rather than a missing key error.

solnic:
Lazy-loading should only be used when container is not finalized. I think eventually we need a way of configuring containers with different loading strategies so that in case of a finalized container (as in, application's runtime environment), we simply expect all objects to be registered already.

timriley:
I updated this since @solnic's last comment. What do you think now? Happy to see it merged?

lib/dry/system/container.rb
```diff
@@ -413,6 +413,13 @@ def root
           config.root
         end
 
+        # @api public
+        def resolve(key)
+          load_component(key) unless frozen?
+
+          super(key)
```
solnic: just `super` :)

solnic:
This is a nice code removal. I only left one minor comment. Let's get this merged and see how it works for us.

PR #42, opened 2016-12-10
Require names to be provided when importing other system containers

timriley:
As discussed with @solnic in gitter about 100 years ago, we decided that having implicit names for imported containers wasn't a good idea. This change makes it so `Dry::System::Container.import` requires you to pass a hash of names/containers.

After making this change, the `name` setting for `Container` wasn't used anywhere, so I removed this too.

### DIFF
--- a/lib/dry/system/component.rb
+++ b/lib/dry/system/component.rb
@@ -95,7 +95,6 @@ module Dry
       # @example
       #   class MyApp < Dry::System::Container
       #     configure do |config|
-      #       config.name = :my_app
       #       config.root = Pathname('/my/app')
       #     end
       #
--- a/lib/dry/system/container.rb
+++ b/lib/dry/system/container.rb
@@ -104,7 +104,6 @@ module Dry
         #   class Core < Dry::System::Container
         #     configure do |config|
         #       config.root = Pathname("/path/to/app")
-        #       config.name = :core
         #       config.auto_register = %w(lib/apis lib/core)
         #     end
         #   end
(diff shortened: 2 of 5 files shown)


timriley:
I'll await your feedback before merging this, @solnic, as removing the `name` setting isn't something we'd discussed before.

solnic:
Are we sure we want to remove `Container.name`? I've got a feeling it'll be useful for stuff like logging and instrumentation, hmm...

timriley:
@solnic and while I'm checking old PRs, ping on this one too. Perhaps the safest thing here is for me to reinstate that container `name` setting (even though we're not using it for anything right now) and then merge this?

If we leave the setting, perhaps we could add an extra method to make it so that configuring a name isn't absolutely necessary, e.g.

```
def self.name
  self.config.name || Dry::Core::Inflections.underscore(name)
end
```

That way if you're happy to just let the class name become the name of the container, you don't need to manually configure the name setting every time you create your own container.

solnic:
I believe this will work:

``` ruby
setting :name, Core::Inflector.underscore(name).to_sym
```

Other than that, I'm good with this change.

timriley:
I'm going to split this into 2 PRs and merge them one by one.

timriley:
> I believe this will work:
> 
> `setting :name, Core::Inflector.underscore(name).to_sym`
>
> Other than that, I'm good with this change.

@solnic, that didn't actually work, because when the default is evaluated, the current scope is the `Dry::System::Container` class definition, so the default becomes `:"dry/system/container"` for everything.

Looking into this further, I think it's a little too dicey for us to determine a default name automatically, since container classes could realistically fall under various module/class naming schemes. Better just to keep things as they are, I think, asking users to provide an explicit `name` config if they want one.

PR #44, opened 2017-03-13
Establish concept of manual registrations in Container (and support lazy loading)

### LINKED ISSUE #33: Manually registered container items aren't lazily loaded

timriley:
This PR makes manual registration files a proper, configurable part of the container setup.

One benefit of this is that we can now attempt to load a matching manual registration file during lazy loading, which, alongside the lazy loading of bootable deps, means there should be no gaps in what kind of things can be lazily loaded during development.

Resolves #33.

@solnic @AMHOL – what do you think? If you're happy with this direction I'll add more tests to ensure everything's covered.

In working through this, I noticed a few structural things I'd like to improve later. I'm not sure if e.g. `Component#bootable?` and `#boot_file?` should exist - it feels like that's the component knowing too much about other parts of the system. Would it be cleaner for this to be provided by something like `Booter#bootable?(component)`? This is the approach I took with `ManualRegistrar#file_exists?`, anyway.

### DIFF
--- a/lib/dry/system/constants.rb
+++ b/lib/dry/system/constants.rb
@@ -1,6 +1,7 @@
 module Dry
   module System
     RB_EXT = '.rb'.freeze
+    RB_GLOB = '*.rb'.freeze
     EMPTY_STRING = ''.freeze
     PATH_SEPARATOR = '/'.freeze
     DEFAULT_SEPARATOR = '.'.freeze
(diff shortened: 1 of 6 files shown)


AMHOL:
I like this approach, feels like service provider in traditional IoC

timriley:
@solnic ping - what were your thoughts on this?

solnic:
I'm super late with this feedback, but:

> In working through this, I noticed a few structural things I'd like to improve later. I'm not sure if e.g. Component#bootable? and #boot_file? should exist - it feels like that's the component knowing too much about other parts of the system.

Yes, I wanted to do the same thing so 👍

PR #45, opened 2017-04-23
Control auto-registration via magic comments

timriley:
Auto-registration is one of my favourite dry-system features, and our usual practice has been to point it at our entire `lib/` folders and have it auto-register everything.

The problem is, there's a lot of files in there (e.g. entity or value classes) that don't need to be available in the container - those files we `require` directly and then instantiate their classes directly.

Now, it's mostly harmless for dry-system to register these, but it'd be nice if it didn't. We only want useful registrations in our containers.

One option would be to move these files into a totally separate directly structure, one which the auto-registrar doesn't look at, but that's a big disturbance to our file structure due to some implementation detail of dry-system. We don't want the tail to wag the dog.

Another option would be to maintain some kind of blacklist/whitelist of files to auto-register, but this would become laborious to maintain and would most likely get out of sync.

The best option, I think, is allowing the auto-registration behaviour to be specified _right there in the source file:_

```ruby
# auto_register: false

class MyValueObject
  def initialize(attrs)
    # ...
  end
end
```

This PR makes a first pass at enabling this behaviour. It's not polished yet, but it works, and I wanted to put it up here now to see what @solnic and @AMHOL think about this :)

A benefit of this approach the configuration is _right there with the code,_ so it should be fairly obvious what the deal is, and you're much less likely to forget about it.

And the bigger win, IMO, would be that in dry-web apps, most of the stuff that we're currently throwing into the system "core" dirs (e.g. `apps/main/core/main`) could go straight into the `lib/` dir with the rest of that system's ruby source files. The core dirs have wound up being pretty cluttered, mostly I think because that's where we want to tuck special classes that we don't want automatically registered in the container. Keeping everything in `lib/`, like a standard Ruby project, I think will make the structure of dry-web apps much easier to understand for new users.

@solnic – does the place I've integrated this feature seem good to you? Let me know if you have any thoughts, code-wise, and I'll be happy to incorporate them :)

### DIFF
--- a/lib/dry/system/auto_registrar.rb
+++ b/lib/dry/system/auto_registrar.rb
@@ -56,11 +56,13 @@ module Dry
         Dir["#{root}/#{dir}/**/*.rb"]
       end
 
+      # @api private
       def relative_path(dir, file_path)
         dir_root = root.join(dir.to_s.split('/')[0])
         file_path.to_s.sub("#{dir_root}/", '').sub(RB_EXT, EMPTY_STRING)
       end
 
+      # @api private
       def file_options(file_name)
         MagicCommentsParser.(file_name)
       end
--- a/lib/dry/system/auto_registrar.rb
+++ b/lib/dry/system/auto_registrar.rb
@@ -46,8 +46,8 @@ module Dry
       # @api private
       def components(dir)
         files(dir).
-          map { |file_path| [file_path, file_options(file_path)] }.
-          map { |(file_path, options)| component(relative_path(dir, file_path), **options) }.
+          map { |file_name| [file_name, file_options(file_name)] }.
+          map { |(file_name, options)| component(relative_path(dir, file_name), **options) }.
           reject { |component| key?(component.identifier) }
       end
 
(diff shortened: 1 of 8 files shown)


solnic:
I think this is OK, definitely the most flexible solution. We may also consider adding a configuration option ie `config.auto_registration_exclude { |dir| dir.include?('entities') }` or something like that. FWIW I'd prefer to configure it than to remember to add `auto_register: false` to every entity class file.

timriley:
Thanks @solnic, I like that idea too. What I might do is finish this magic comments support and then move onto something like that in a separate PR.

solnic:
@timriley yes this sounds reasonable

timriley:
Alrighty, @solnic, what do you reckon, good to merge in now?

I guess the only thing left to consider might be the performance of the auto-registrar now that its reading the top few lines of the files looking for magic comments.

solnic:
I tested it already and there's no perf impact, hard to say how fast/slow it'll be *if* you actually use these comments, but we'll find out soon enough

alexandru-calinoiu:
I think this should be mentioned somewhere in the docs.

I can't find the repo for `http://dry-rb.org/gems/dry-system/auto-import/` to send a pr.

solnic:
@alexandru-calinoiu it's https://github.com/dry-rb/dry-rb.org

Issue #46, opened 2017-05-16
Support skipping auto-registration for particular files via configuration code


timriley:
In https://github.com/dry-rb/dry-system/pull/45#issuecomment-296590276, @solnic, you mentioned this:

> We may also consider adding a configuration option ie config.auto_registration_exclude { |dir| dir.include?('entities') } or something like that.

I'm looking to start adding this now, so it can complement the magic comments behaviour and then we can consider a new dry-system release.

However, I feel hesitant to add another setting like `config.auto_registration_exclude`, for a few reasons:

1. Firstly, dry-configurable won't make it easy to have a block as the value for the setting like your example. Since it uses blocks for nested settings, we'd have to ask users to provide a proc as the value, which doesn't feel as natural
2. I'm also wary about "configuration-creep". If we can avoid a setting at the top-level of the container, I think that'd be good.
3. The user may not want this setting to apply to all their auto-registration paths.
4. Lastly, and perhaps most critically, the auto-registrar class is a setting _itself_, and if the default auto-registrar was replaced, it may not even want to pay any attention to a setting like `auto_registration_exclude`.

So instead of adding another container setting, I wonder if we could reuse the `.auto_register!(path, &block)` method we alreay have, just extending the block so it supports auto-registration configuration for that particular path. So instead of:

```ruby
auto_register!('lib') do |component|
  # some custom initialization logic
  constant = component.loader.constant
  constant.create
end
```

We make it this:

```ruby
auto_register!('lib') do |config|
  # Both of these configuration blocks are optional

  config.instance do |component|
    # some custom initialization logic
    constant = component.loader.constant
    constant.create
  end

  config.exclude do |component|
    component.path =~ /entities/
  end
end
```

What do you think? Since this configuration is coupled to our default `Dry::System::AutoRegistrar`, it feels like a more appropriate place for it.

solnic:
@timriley oh this is a really nice idea!

timriley:
Resolved by #47

PR #47, opened 2017-06-02
Skip auto registration via config

GustavoCaso:
This implements https://github.com/dry-rb/dry-system/issues/46

@timriley is we mention this add the ability to skip `auto_registration` via configuration.

### DIFF
--- a/examples/custom_configuration_auto_register/system/container.rb
+++ b/examples/custom_configuration_auto_register/system/container.rb
@@ -1,0 +1,16 @@
+require 'dry/system/container'
+
+class App < Dry::System::Container
+  load_paths!('lib', 'system')
+
+  auto_register!('lib') do |config|
+    config.instance do |component|
+      # some custom initialization logic
+      component.instance
+    end
+
+    config.exclude do |component|
+      component.path =~ /entities/
+    end
+  end
+end
--- a/lib/dry/system/auto_registrar.rb
+++ b/lib/dry/system/auto_registrar.rb
@@ -1,5 +1,6 @@
 require 'dry/system/constants'
 require 'dry/system/magic_comments_parser'
+require 'dry/system/auto_registrar/configuration'
 
 module Dry
   module System
@@ -27,16 +28,14 @@ module Dry
       end
 
       # @api private
-      def call(dir, &block)
+      def call(dir)
+        registration_config = Configuration.new
+        yield(registration_config) if block_given?
         components(dir).each do |component|
-          next if !component.auto_register?
+          next if !component.auto_register? || registration_config.exclude.(component)
 
           container.require_component(component) do
-            if block
-              register(component.identifier, yield(component))
-            else
-              register(component.identifier) { component.instance }
-            end
+            register(component.identifier) { registration_config.instance.(component) }
           end
         end
       end
--- a/lib/dry/system/container.rb
+++ b/lib/dry/system/container.rb
@@ -316,7 +316,8 @@ module Dry
         #
         # Typically you want to configure auto_register directories, and it will
         # work automatically. Use this method in cases where you want to have an
-        # explicit way where some components are auto-registered.
+        # explicit way where some components are auto-registered, or if you want
+        # to exclude some components from been auto-registered
         #
         # @example
         #   class MyApp < Dry::System::Container
@@ -328,15 +329,21 @@ module Dry
         #     auto_register!('lib/core')
         #
         #     # with a dir and a custom registration block
-        #     auto_register!('lib/core') do |component|
-        #       # custom way of initializing a component
+        #     auto_register!('lib/core') do |config|
+        #       config.instance do |component|
+        #         # custom way of initializing a component
+        #       end
+        #
+        #       config.exclude do |component|
+        #         # return true to exclude component from auto-registration
+        #       end
         #     end
         #   end
         #
         # @param [String] dir The dir name relative to the root dir
         #
-        # @yield [Component]
-        # @see [Component]
+        # @yield AutoRegistrar::Configuration
+        # @see AutoRegistrar::Configuration
         #
         # @return [self]
         #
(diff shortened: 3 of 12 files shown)


lib/dry/system/auto_registrar.rb
```diff
@@ -28,12 +29,14 @@ def finalize!
 
       # @api private
       def call(dir, &block)
+        config = AutoRegistraionConfig.new
```
GustavoCaso: For this part of the code I had different solutions in mind I wanted to share with you to see which one fits more with your idea. I will try to explain what the other idea I had.

1.- The one in the PR, we leave the `config` instance the logic to decide if the proc object for `instance` or `exclude` has been store, one thing that I do not like is that we always instanciate `AutoRegistraionConfig ` even when there is no block passed

2.- The other was this one 
```ruby
def call(dir, &block)
   yield(config = AutoRegistraionConfig.new) if block_given?
   components(dir).each do |component|
      next if !component.auto_register? || (config && config.call(:exclude, component))

  container.require_component(component) do
    if initialized_component = config && config.call(:instance, component)
      register(component.identifier) { initialized_component }
    else
    end
  end
 end
```

This solution avoid creating unneeded instance of `config` but since there is chance `config` is `nil` we have to double check, not really fun of this one

3.- The other solution would instantiate the configuration in the initializer of `auto_register`

```ruby
def initialize(container)
  @container = container
  @config = container.config
  @auto_registration_config = AutoRegistraionConfig.new
end

# @api private
def call(dir, &block)
  yield(auto_registration_config) if block_given?
  components(dir).each do |component|
    next if !component.auto_register? || auto_registration_config.call(:exclude, component)

    container.require_component(component) do
       if initialized_component = auto_registration_config.call(:instance, component)
         register(component.identifier) { initialized_component }
       else
       end
    end
  end
end
```

timriley:
This 

Regarding your https://github.com/dry-rb/dry-system/pull/47/commits/4e8bb37b190e873aa6bf1ae8b9a1cb5d50956713, I don't think we actually want this. A user may want to provide different auto-registration behaviour for different paths, so we _do_ want a new configuration object to be built  with each `#call`.

I'm also not sure we want to keep https://github.com/dry-rb/dry-system/pull/47/commits/fe7e4072d8abe517487006ea179053b153de84ef, where you used a `Struct` as the basis for the configuration class. Structs offer a lot of other behaviour that we don't really want in this situation. I'm happy for it to be a regular class with a couple of special accessors.

Something like this, maybe?

```ruby
class Config
  DEFAULT_INSTANCE = -> component { component.instance }.freeze
  FALSE_PROC = -> * { false }.freeze

  def self.setting(name)
    define_method(name) do |&block|
      ivar = "@#{name}"

      if block
        instance_variable_set(ivar, block)
      else
        instance_variable_get(ivar)
      end
    end
  end

  setting :exclude
  setting :instance

  def initialize
    @instance = DEFAULT_INSTANCE
    @exclude = FALSE_PROC
  end
end
```

lib/dry/system/auto_registration_config.rb
```diff
@@ -0,0 +1,22 @@
...
+        @exclude = block
+      end
+
+      def call(method, component)
+        if proc = instance_variable_get("@#{method}")
+          proc.(component)
+        else
+          nil
+        end
+      end
```
timriley: If we went along the lines of the example I left you in my last comment, I don't think we'd need this `#call`. Instead, we could just send the `#instance` or `#exclude` messages directly, e.g. `registration_config.exclude.(component)`, which feels a little nicer.

lib/dry/system/auto_registration_config.rb
```diff
@@ -0,0 +1,22 @@
+module Dry
+  module System
+    # @api private
+    AutoRegistraionConfig = Struct.new(:instance, :exclude) do
```
timriley: Perhaps we could rename this to `Dry::System::AutoRegistrar::Configuration`? We typically spell the `configuration` word out in full for classes like this. And by namespacing it, we make it clear that the configuration relates to the `AutoRegistrar` only, and it means we can just refer to its  constant as `Configuration` from within the `AutoRegistrar` class.

solnic:
Please rebase/squash before merging.

GustavoCaso:
Rebased and squashed

examples/custom_configuration_auto_register/lib/entities.rb
```diff
@@ -0,0 +1,6 @@
+require 'import'
+
+class Entities
+  include Import['persistence.db']
+
+end
```
timriley: It'd be more representative of a real app if we had `lib/entities/user.rb` instead of `lib/entities.rb`. Would you mind doing that?

examples/custom_configuration_auto_register/system/container.rb
```diff
@@ -0,0 +1,19 @@
+require 'dry/system/container'
+
+class App < Dry::System::Container
+  configure do |config|
+  end
```
timriley: We can cut this out from this example if we're not configuring anything.

lib/dry/system/auto_registrar.rb
```diff
@@ -28,15 +29,13 @@ def finalize!
 
       # @api private
       def call(dir, &block)
```
timriley: We don't actually need this `&block` arg since we're just yielding the config.

lib/dry/system/container.rb
```diff
@@ -328,15 +329,21 @@ def load_paths!(*dirs)
         #     auto_register!('lib/core')
         #
         #     # with a dir and a custom registration block
-        #     auto_register!('lib/core') do |component|
-        #       # custom way of initializing a component
+        #     auto_register!('lib/core') do |config|
+        #       config.instance do |component|
+        #         custom way of initializing a component
```
timriley: Need a comment before custom (e.g. `# custom`)

lib/dry/system/container.rb
```diff
@@ -328,15 +329,21 @@ def load_paths!(*dirs)
...
+        #       end
+        #
+        #       config.exclude do |component|
+        #         custom way for excluding components
```
timriley: Would probably be best to read `# return true for components to exclude from auto-registration`

timriley:
Great stuff, thanks @GustavoCaso! (And thanks @solnic for taking a look too)

Issue #49, opened 2017-06-05
Add test mode as a first-class concept


solnic:
We need a way to tell the system that it's in test mode and components might be stubbed so we can't freeze a container upon finalization.

refs #32

PR #55, opened 2017-06-10
Added enabled_stubs! for testing finalize! containers

GustavoCaso:
Solves #49 

After giving some thoughts and looking at how we implement stubbing in `dry-container` I came up with this solution.

There a few things I do not like is that the `finalize!` method in the `Stub` module is practically the same except for the `freeze` part.

I had other implementation where we pass an argument to `finalize!` method making `freeze` control by this arguments.

```ruby
def finalize!(testing: false, &block)
  return self if frozen?

  yield(self) if block

  importer.finalize!
  booter.finalize!
  manual_registrar.finalize!
  auto_registrar.finalize!

  freeze unless testing
end
```

And inside the `Stub` module we can simply.
```ruby
def finalize!(&block)
  super(testing: true, &block)
end
```

Or even in test the user can say `finalize(testing: true)` without the need of the `Stub` module at all

@solnic also point out that we could bring back the `env` setting back to `dry-system`

### DIFF
--- a/lib/dry/system/container.rb
+++ b/lib/dry/system/container.rb
@@ -239,7 +239,7 @@ module Dry
         # @return [self] frozen container
         #
         # @api public
-        def finalize!(&block)
+        def finalize!(freezable: true, &block)
           return self if frozen?
 
           yield(self) if block
@@ -249,7 +249,7 @@ module Dry
           manual_registrar.finalize!
           auto_registrar.finalize!
 
-          freeze
+          freeze if freezable
         end
 
         # Boots a specific component
--- a/lib/dry/system/container.rb
+++ b/lib/dry/system/container.rb
@@ -239,7 +239,7 @@ module Dry
         # @return [self] frozen container
         #
         # @api public
-        def finalize!(freezable: true, &block)
+        def finalize!(freeze: true, &block)
           return self if frozen?
 
           yield(self) if block
@@ -249,7 +249,7 @@ module Dry
           manual_registrar.finalize!
           auto_registrar.finalize!
 
-          freeze if freezable
+          self.freeze if freeze
         end
 
         # Boots a specific component
--- a/lib/dry/system/stub.rb
+++ b/lib/dry/system/stub.rb
@@ -1,0 +1,27 @@
+module Dry
+  module System
+    class Container
+      # Incuded only in the Test environment
+      # Sending the message enabled_stubs! allow you to stub finalize container
+      # in your tests.
+      #
+      # @api private
+      module Stub
+        def finalize!(&block)
+          yield(self) if block
+
+          importer.finalize!
+          booter.finalize!
+          manual_registrar.finalize!
+          auto_registrar.finalize!
+        end
+      end
+
+      class << self
+        def enabled_stubs!
+          extend ::Dry::System::Container::Stub
+        end
+      end
+    end
+  end
+end
--- a/lib/dry/system/stub.rb
+++ b/lib/dry/system/stub.rb
@@ -2,8 +2,8 @@ module Dry
   module System
     class Container
       # Incuded only in the Test environment
-      # Sending the message enabled_stubs! allow you to stub finalize container
-      # in your tests.
+      # Sending the message enable_stubs! allow you to stub components after
+      # finalize your container in your tests.
       #
       # @api private
       module Stub
@@ -18,7 +18,7 @@ module Dry
       end
 
       class << self
-        def enabled_stubs!
+        def enable_stubs!
           extend ::Dry::System::Container::Stub
         end
       end
--- a/lib/dry/system/stubs.rb
+++ b/lib/dry/system/stubs.rb
@@ -6,7 +6,7 @@ module Dry
       # finalize your container in your tests.
       #
       # @api private
-      module Stub
+      module Stubs
         def finalize!(&block)
           yield(self) if block
 
@@ -19,7 +19,7 @@ module Dry
 
       class << self
         def enable_stubs!
-          extend ::Dry::System::Container::Stub
+          extend ::Dry::System::Container::Stubs
         end
       end
     end
--- a/lib/dry/system/stubs.rb
+++ b/lib/dry/system/stubs.rb
@@ -8,12 +8,7 @@ module Dry
       # @api private
       module Stubs
         def finalize!(&block)
-          yield(self) if block
-
-          importer.finalize!
-          booter.finalize!
-          manual_registrar.finalize!
-          auto_registrar.finalize!
+          super(freezable: false, &block)
         end
       end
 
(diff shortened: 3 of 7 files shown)


AMHOL:
@GustavoCaso nice work! :smile:

I think this should be `enable_stubs!` rather than `enabled_stubs!`

GustavoCaso:
@AMHOL you are absolutely right. I will change it 😄

AMHOL:
Nice one @GustavoCaso

AMHOL:
Looks good for merge to me, @solnic @timriley  ?

lib/dry/system/stub.rb
```diff
@@ -0,0 +1,27 @@
+module Dry
+  module System
+    class Container
+      # Incuded only in the Test environment
+      # Sending the message enable_stubs! allow you to stub components after
+      # finalize your container in your tests.
+      #
+      # @api private
+      module Stub
```
timriley: Feels to me like this could be called `Stubs` rather than just `Stub`? That feels more natural to me since it's about a stubbing _capability_ in general, not just one stub )

lib/dry/system/stub.rb
```diff
@@ -0,0 +1,27 @@
...
+        def finalize!(&block)
+          yield(self) if block
+
+          importer.finalize!
+          booter.finalize!
+          manual_registrar.finalize!
+          auto_registrar.finalize!
```
timriley: Your point about this method duplicating al the calls of the main `.finalize!` method is a good one.

Feels like it'd be _ok_ if we added a `freeze:` boolean kwarg to the main `.finalize!` and just pass `super(freeze: true, &block)` like you mention.  I think we'd want to make that kwarg part of our "private" API though, since we'd want people to use `enable_stubs!` instead of manually passing `freeze: true`.

timriley:
I have a question on whether we merge this feature now or hold it back for bringing in an `env` of some kind:

In the near-term, would our different "environments" for dry-system change any behaviour _apart_ from just whether the container is frozen or not? If not, then we could probably just merge this and think about adding environments later.

I'm sure there are interesting things we could do with environments in the longer term, like _maybe_ having the container help with code reloading in dev, or giving the user some extra env hooks so they can add their own custom boot logic for different environments.

I just want to make sure we consider our broader plans before committing to a path that might lead to unnecessary API churn :)

GustavoCaso:
@timriley I have applied the desire changes. One thing the keyword argument `freeze` was messing with the actual `freeze` method. Please feel free if you do not like this name just mention it

flash-gordon:
@GustavoCaso you can work around that with `self.freeze if freeze`

GustavoCaso:
@timriley @solnic what do you think the final code?

lib/dry/system/container.rb
```diff
@@ -249,7 +249,7 @@ def finalize!(&block)
           manual_registrar.finalize!
           auto_registrar.finalize!
 
-          freeze
+          self.freeze if freeze
```
solnic: this will return `nil` if it's frozen already, so it needs to be `freeze ? self.freeze : self`

lib/dry/system/stubs.rb
```diff
@@ -0,0 +1,22 @@
...
+        def finalize!(&block)
+          super(freeze: false, &block)
+        end
+      end
+
+      class << self
+        def enable_stubs!
+          extend ::Dry::System::Container::Stubs
+        end
+      end
```
solnic: `class << self` feels odd here, just add extend to the re-opened `Container` class

spec/fixtures/stubbing/system/boot/mock.rb
```diff
@@ -0,0 +1,4 @@
+Test::Container.finalize(:mock) do |container|
+
```
solnic: Please remove this blank line

spec/unit/container_spec.rb
```diff
@@ -195,4 +195,49 @@ class Test::Container < Dry::System::Container
...
+      class Test::Container < Dry::System::Container
+        configure do |config|
+          config.root = SPEC_ROOT.join('fixtures/stubbing').realpath
+        end
+        load_paths!('lib')
+        auto_register!('lib')
+      end
+    end
+
+
```
solnic: Please remove this extra empty line

spec/unit/container_spec.rb
```diff
@@ -195,4 +195,49 @@ class Test::Container < Dry::System::Container
...
+      end
+    end
+
+    describe 'with enable_stubs!' do
+      before do
+        container.enable_stubs!
+        container.finalize!
+      end
+
+      it 'allow to stub the container it self' do
```
solnic: Should be `it 'allows ...'` (here and in the next one)

lib/dry/system/container.rb
GustavoCaso: I think the first line of the method takes that into consideration `return self if frozen?`

lib/dry/system/stubs.rb
GustavoCaso: But how are going the user tell when the want to enable the stubs?

lib/dry/system/container.rb
solnic: Oh I missed that. Damn early returns ;) ignore my comment then

lib/dry/system/stubs.rb
solnic: Gah I meant that you can drop `class << self` and simply do `def self.enable_stubs!` which does the extend.

lib/dry/system/container.rb
AMHOL: Can `freeze` not be `false` if its not already frozen?

GustavoCaso:
@solnic @timriley any more concerns regarding this PR?

timriley:
I reckon this'd be fine to pull in now. The thoughts I shared earlier were about whether we'd want to follow a different design for this feature if we were to adopt the idea of "environments" within dry-system in the near future.

But I think this is a useful feature on its own and the disturbance later on if we introduced a broader "environments" feature wouldn't be too great.

So let's get it in. @solnic — you happy with this?

GustavoCaso:
Thanks ❤️

solnic:
Seems like I didn't pay enough attention when looking at this PR. This is not working at all, unfortunately, and I realised this after releasing 0.7.2 :( I'll fix it and push 0.7.3 ASAP.

solnic:
OK I fixed it https://github.com/dry-rb/dry-system/commit/19495cdb0543ef3ed00d5253925947e64c38e94c and https://github.com/dry-rb/dry-system/commit/5d7fac6bdef22bc7579ce4a6ae2d75e62c866e60

In the future we should really test things out in the real world. dry-system is not a big lib, but it is complex, and some times it's hard to tell if something truly works, despite having specs written :)

Issue #50, opened 2017-06-05
Allow same key in a component path to occur more than once


solnic:
Refs #41

GustavoCaso:
I was solve #58

PR #58, opened 2017-06-20
Allow same key two occur multiple times in path

GustavoCaso:
Solves #50 

Apart from allowing the same in the path, this change raises and exception if the namespace set up for the component is not found in the path


solnic:
This is very helpful, thank you Gustavo :)

PR #59, opened 2017-08-28
3rd party bootable components

solnic:
This adds support for external components. To enable this we need to add a couple of new concepts, like a component provider and external component. This is still a WIP, but I got things figured out, so I'll clean this up and should be ready to go.

### API

``` ruby
# ie in a gem you can do this to register your provider
require "dry/system"

Dry::System.register_provider(
  :external_components,
  boot_path: Pathname(__dir__).join('../components').realpath
)

# then each component file can register something
# ie `../components/logger.rb'
require "dry/system"

Dry::System.register_component(:logger, provider: :external_components) do
  init do
    module ExternalComponents
      class Logger
      end
    end
  end

  start do
    register(:logger, ExternalComponents::Logger.new)
  end
end

# then in your container you can ask for an external logger:

# let's assume there's a gem called like that
require "external_components"

  register_external(:logger, provider: :external_components)
end

# you can use a custom registration key
  register_external(:logger, provider: :external_components, as: "core.logger")
end

# you can hook into its lifecycle too:
MyContainer.on(logger: :init) do |step|
  # here it's guaranteed that component init is done
end
```

### TODO

- [x] Add ability to register an external component
- [x] Add ability to register a component provider
- [x] Add ability to register a component using external identifier
- [x] Add ability to register components with the same identifiers under custom keys in the container
- [x] Add ability to select which provider should be used when registering an external component
- [x] Clean things up by introducing explicit `Provider` and `ExternalComponent` concepts (they should encapsulate all the required information, so that we don't have to pass many things around)

### DIFF
--- a/lib/dry/system.rb
+++ b/lib/dry/system.rb
@@ -1,4 +1,27 @@
+require 'dry/system/provider'
+require 'dry/system/provider_registry'
+
 module Dry
   module System
+    # Register external component provider
+    #
+    # @api public
+    def self.register_provider(identifier, options)
+      providers.register(identifier, options)
+      self
+    end
+
+    # Register an external component that can be booted within other systems
+    #
+    # @api public
+    def self.register_component(identifier, provider:, &block)
+      providers[provider].register_component(identifier, block)
+      self
+    end
+
+    # @api private
+    def self.providers
+      @__providers__ ||= ProviderRegistry.new
+    end
   end
 end
--- a/lib/dry/system/booter.rb
+++ b/lib/dry/system/booter.rb
@@ -1,3 +1,4 @@
+require 'dry/system/components/bootable'
 require 'dry/system/errors'
 require 'dry/system/lifecycle'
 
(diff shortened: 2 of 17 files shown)


solnic:
I'll add support for passing settings to components in another PR. I'll also make things more robust in master directly (it's missing proper error handling).

timriley:
@solnic Thanks for getting this started!

What was your thinking behind organising 3rd party components within "providers"?

A simpler alternative would be merely exposing a constant (e.g. `Alternative::LoggerComponent`) and having that be a thing that can be registered within a system, so I'd just be interested to here your thoughts on what extra facilities or conveniences the `registered provider->registered component` structure would bring.

(One thing that comes to mind is that by avoiding constants for 3rd party components is that we don't have to name them awkwardly, e.g. with `Alternative::Logger` being the logger class and then having to provide `Alternative::Component::Logger` or `Alternative::LoggerComponent` for the registrable component).

solnic:
@timriley because it's a distinct concept and it's good to have it represented explicitly. Notice that providers have information about boot paths, boot files and so on. Otherwise we'd have external components holding a lot of knowledge (and it'd be duplicated in cases where multiple components come from the same source).

In general, every time we didn't have explicit representation of some concept in this gem, things got weirdly complicated. I think we'll understand this feature better when we actually try to use it in real world, it's possible we'll see that it can be simplified, not sure :)

PR #61, opened 2017-10-09
Improved external components

solnic:
Refactor bootable components and improve `boot` feature.

This is a substantial improvement, I managed to unify bootable/external components, so now we only have `Components::Bootable` and it doesn't matter if it's exernal or internal. Here's an overview of what this PR allows:

### New `boot` syntax

We now have `Container.boot` and `finalize` was deprecated. It supports additional `namespace` option which is cool, because extra nesting is no longer needed:

``` ruby
# default behavior is registration w/o namespace
MyApp::Container.boot(:db) do
  init do
    require "some_gem/db"
  end

  start do
    register(:conn, SomeGem::Db.new)
  end
end

MyApp::Container["conn"]

# namespacing is simpler now
MyApp::Container.boot(:db, namespace: true) do
  init do
    require "some_gem/db"
  end

  start do
    register(:conn, SomeGem::Db.new)
  end
end

# you can provide a custom namespace too
MyApp::Container.boot(:db, namespace: "backend.persistence") do
  init do
    require "some_gem/db"
  end

  start do
    register(:conn, SomeGem::Db.new)
  end
end

MyApp::Container["backend.persistence.conn"]
```

## External components are just bootable components

This means you register them using standard `Container.boot` API and provide `:from` option which specifies the component provider:

``` ruby
MyApp::Container.boot(:persistence, namespace: true, from: :some_provider) do
  after(:start) do |c|
    register(:conn, c[:database_connection])
  end
end

MyApp::Container['persistence.conn']
```

## Configuration API is accessible via `boot`

Because there's no distinction between external/internal bootable components, the new `configure` API is now always available:

``` ruby
# if the provider sets up the component like this:
Dry::System.register_component(:persistence, provider: :some_provider) do
  start do
    register(:database_connection, SomeProvider::DB.new(config.database_url)
  end
end

# then this configuration in your app can be set like this
MyApp::Container.boot(:persistence, namespace: true, from: :some_provider) do
  configure do |config|
     config.database_url = "sqlite::memory"
  end

  # NOTICE that hook blocks receive external component instance where
  # objects are provided, so that you can register them under your identifiers
  after(:start) do |c|
    register(:conn, c[:database_connection])
  end
end

# and this works too using internal components
MyApp::Container.boot(:persistence, namespace: true) do
  configure do |config|
     config.database_url = "sqlite::memory"
  end

  start do
    register(:conn, MyDB.new(config.database_url)
  end
end
```

### Bootable components are now accessible, standalone objects

This won't be super popular, but it can be helpful. Containers register their bootable components under `Container.components`.

Here's a sample pry session in an actual app, where we only require container file and then manually mess with a component:

``` ruby
[1] pry(main)> require App::Container.booter.path.join("persistence")
=> true
[2] pry(main)> p = App::Container.components[:persistence]
=> #<Dry::System::Components::Bootable:0x00007fdd95228dd0
 @identifier=:persistence,
 @namespace=true,
 @options={:container=>App::Container, :namespace=>true, :block=>#<Proc:0x00007fdd95228e98@/Users/solnic/Workspace/test-app/system/boot/persistence.rb:1>},
 @triggers={:before=>{}, :after=>{}}>
[3] pry(main)> p.init
=> #<Dry::System::Components::Bootable:0x00007fdd95228dd0
 @identifier=:persistence,
 @lifecycle=Kernel,
 @namespace=true,
 @options={:container=>App::Container, :namespace=>true, :block=>#<Proc:0x00007fdd95228e98@/Users/solnic/Workspace/test-app/system/boot/persistence.rb:1>},
 @triggers={:before=>{}, :after=>{:init=>[]}}>
[4] pry(main)> p.statuses
=> [:init]
[5] pry(main)> p.start
=> #<Dry::System::Components::Bootable:0x00007fdd95228dd0
 @identifier=:persistence,
 @lifecycle=Kernel,
 @namespace=true,
 @options={:container=>App::Container, :namespace=>true, :block=>#<Proc:0x00007fdd95228e98@/Users/solnic/Workspace/test-app/system/boot/persistence.rb:1>},
 @triggers={:before=>{}, :after=>{:init=>[], :start=>[]}}>
[6] pry(main)> p.statuses
=> [:init, :start]
```

We can add more convenient methods later on, what's important is that we now have nice access to these objects.

## In-line booting is now supported

It is now possible to define bootable components w/o having special files. This means that you can just do this in your container class:

``` ruby
  # a local one (trivial example I know, just showing that it's possible :P)
  boot(:persistence, namespace: true) do
    start do
      register(:conn, Db.new)
    end
  end

  # an external local one
  boot(:persistence, namespace: true, from: :some_provider) do
    configure do |config|
       config.database_url = "sqlite::memory"
    end

    after(:start) do |c|
      register(:conn, c[:database_connection])
    end
  end
end
```

It will work just like with boot files, this is really cool as it could be used to set up lightweight applications w/o complex dir structures and many files, yet, it would benefit from isolated requires/booting process.

## Lifecycle triggers / hooks and execution context

There are a couple things to know:

* `init`, `start` and `stop` triggers are executed in the context of a lifecycle object. It has access to target container (as in, your application container), and it registers objects inside **its temporary container**, then we copy whatever was registered to the target container.
* `after` and `before` hooks are executed in the context **of your container** so that you can easily call `register` which makes the syntax less verbose and you have access to common methods like `root` etc. In addition, these blocks *receive the container that was used by lifecycle object*, so that you can manually copy registered objects and register them under your keys.
* `boot` block receives *your container* in case you want to refer to it, but it shouldn't really be needed so try to avoid it


lib/dry/system/components/external.rb
```diff
@@ -3,10 +3,68 @@
...
+          triggers[:after][event] << block
+          self
+        end
+
         def external?
           true
         end
+
+        def configure(&block)
+          @config = Config.new do |config|
```
AMHOL: Might be missing something but I think this could just be `Config.new(&block)`?

solnic: @AMHOL fixed, thanks (:

solnic:
Just wanted to mention that when I worked on this, I realized **eventually** we should end up with just bootable components, without special-casing "external" ones. After all an application **is a component provider** too, so there's no real difference between an app, and some external provider. Both have boot files under specified boot path, both register components to be finalized within a given container. For now I kept it separate as it'd be too much to change at once, but I'll definitely look into unifying this before 1.0.0.

spec/fixtures/external_components/components/logger.rb
```diff
@@ -7,11 +7,26 @@ class Logger
...
   end
 
   start do
-    register(:logger, ExternalComponents::Logger.new)
+    logger =
+      if config
+        ExternalComponents::Logger.new(config.log_level)
+      else
+        ExternalComponents::Logger.new
+      end
```
timriley: I'm wondering if we should try and move to a slightly more formal configuration structure, either here in this PR or in another before we release this.

I feel like it would be good to have declared configuration keys along with defaults. It'd let external bootable component authors expose a lot more things in config and avoid having these if/else things throughout the component's own lifecycle code.

e.g. if we declared something like `setting :log_level, :info`, the logger could be set up like `ExternalComponents::Logger.new(config.log_level)` regardless of whether the application author provided a custom log level or not.

solnic: @timriley yeah I agree, I'll improve this in a separate PR. I want to port dry-web's settings to dry-system and somehow re-use it for component configuration too

lib/dry/system/lifecycle.rb
```diff
@@ -57,6 +57,25 @@ def call(*triggers)
       end
 
       # @api private
+      def provides?(component)
+        if component.external?
+          container.source != container && container.source.key?(component.identifier)
```
timriley: This `container.source != container` confuses me a little. Looking through `LifecycleContainer`, would't `container.source` _always_ `!=` to any other container, because we always create a branch new `Dry::Container.new` whenever a new `LifecycleContainer` is initialised?

timriley: Also, there's a few places like this where we have to dig through `container` to reach `.source` and ask it questions, like we do here with `container.source.key?`. I'm wondering if we could do a few things to make this a bit better?

- firstly, calling these vars `container` confused me a number of times, because this is different to the normal "container" that dry-system exposes to the outside (which has no `.source` attribute). Perhaps we could pick a better local var name to help distinguish which is which?
- secondly, I wonder if `LifecycleContainer` could offer some extra methods on its own to make it so we don't have to be aware of its internals (i.e. `.source`) whenever we interact with it?

solnic: @timriley yeah it's messy, I want to refactor this once I test things out in a real app

lib/dry/system/container.rb
```diff
@@ -210,10 +211,44 @@ def import(other)
...
+            raise DuplicatedComponentKeyError, "Bootable component #{name.inspect} was already registered"
+          end
+
+          component =
+            if opts[:from]
+              boot_external(name, opts, &block)
+            else
+              boot_local(name, opts, &block)
+            end
+          self
```
GustavoCaso: I think this `self` is not necessary here

lib/dry/system/components/bootable.rb
```diff
@@ -1,37 +1,111 @@
...
+        end
+
+        def stop
+          lifecycle.(:stop)
+          self
+        end
+
+        def finalize
+          lifecycle.container.each do |key, item|
+            container.register(key, item) unless container.key?(key)
```
timriley: I wonder if we should print a warning if there's a conflicting key? Or otherwise we could risk having components that are not fully registered in the main app container?

solnic: @timriley good point, I now wonder if this should raise an exception, it probably should

lib/dry/system/container.rb
solnic: Good point. I'll remove this.

lib/dry/system/components/bootable.rb
solnic: @timriley just realized we do want this check, it is possible that a component was already registered and it's actually fine. See `external_components_spec.rb` where we boot `:logger` from external provider 2 times.

solnic:
This is merged. I'll add docs and maybe clean up some things a little bit in master. FYI this was tested in our apps at Icelab, so I'll release 0.8.0 as we'd like to start using it right away.

PR #62, opened 2017-10-14
Add built-in :system_components with :settings component

solnic:
This introduces built-in `:system_components` with the very first one called `:settings`. It also adds `settings` DSL for bootable components where you can define structure of your settings with setting keys and their types (and it enforces strict types).

Example usage:

``` ruby
class MyApp < Dry::System::Container
  boot :settings, from: :system_components do
    settings do
      key :database_url, type(String).constrained(filled: true)
    end
  end
end

# this uses ENV and .env and .env.#{config.env} to pull in values
MyApp[:settings].database_url
```

Bootable components can ship with their settings predefined too:

``` ruby
Dry::System.register_component(:logger, provider: :some_provider) do
  setttings do
    key :output, type('Any').default($stdout)
    key :level, type(Symbol).default(:debug)
  end

  start do
    require 'some_provider/logger'

    register(:logger, Logger.new(config.output, level: config.level)
  end
end

# then you can boot it and customize:
boot(:logger, from: :some_provider) do
  configure do |c|
    c.level = :error
    c.output = container.root.join('log/app.log')
  end
end
```

### DIFF
--- a/lib/dry/system/components.rb
+++ b/lib/dry/system/components.rb
@@ -1,0 +1,6 @@
+require 'dry/system'
+
+Dry::System.register_provider(
+  :system_components,
+  boot_path: Pathname(__dir__).join('system_components').realpath
+)
--- a/lib/dry/system/settings.rb
+++ b/lib/dry/system/settings.rb
@@ -1,0 +1,67 @@
+require "dry/core/class_builder"
+require "dry/types"
+require "dry/struct"
+
+require "dry/system/settings/file_loader"
+
+module Dry
+  module System
+    module Settings
+      module Types
+        include Dry::Types.module
+      end
+
+      class DSL < BasicObject
+        attr_reader :identifier
+
+        attr_reader :schema
+
+        def initialize(identifier, &block)
+          @identifier = identifier
+          @schema = {}
+          instance_eval(&block)
+        end
+
+        def call
+          Core::ClassBuilder.new(name: 'Configuration', parent: Settings::Configuration).call do |klass|
+            schema.each do |key, type|
+              klass.setting(key, type)
+            end
+          end
+        end
+
+        def key(name, type)
+          schema[name] = type
+        end
+
+        def type(id)
+          Types::Strict.const_get(id.to_s)
+        end
+      end
+
+      class Configuration < Dry::Struct
+        constructor_type :strict_with_defaults
+
+        def self.setting(*args)
+          attribute(*args)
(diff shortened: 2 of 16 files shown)


lib/dry/system/settings.rb
```diff
@@ -0,0 +1,67 @@
...
+            end
+          end
+        end
+
+        def key(name, type)
+          schema[name] = type
+        end
+
+        def type(id)
+          Types::Strict.const_get(id.to_s)
```
timriley: This is pretty neat, but I've a couple of thoughts:

1. We may want to reconsider whether we make these all strict by default, or if we can add some extra layer of coercion we run things through. When we're getting our settings from the .env files or the actual ENV (e.g. in production), they're all strings. In our apps in the past we've actually used some `Form` types (like `Types::Form::Bool`) to coerce these into proper Ruby values.

2. Is hiding the types behind this `type` method actually going to make things more confusing, i.e. make it not really clear that we're directly using dry-types type objects? e.g. `type(String)`  could easily be mistaken for some special API that expects the `::String` constant to be passed to it. Would it be better to just have the settings refer to e.g. `Types::Strict::String` directly?

lib/dry/system/components.rb
```diff
@@ -0,0 +1,6 @@
+require 'dry/system'
+
+Dry::System.register_provider(
+  :system_components,
```
timriley: I wonder if we could just call this `:system`? Perhaps we could make that our standard name for any system-provided components.

timriley:
@solnic I think I must've reviewed this around the same time you merged it. Did you see my comments?

lib/dry/system/settings.rb
solnic: @timriley I think it's valuable to have them strict by default. I'll add custom types for coercions too (I'm aware that we're using `Form::Bool` in some places).

> Would it be better to just have the settings refer to e.g. Types::Strict::String directly?

Right, I thought about this. I guess I can make `Types::*` work and people can decide which types they want to use.

timriley: @solnic OK, this sounds like a good plan then (both the support for easy coercions + ability to use types directly). And yeah, I agree that the strictness is a benefit here.

lib/dry/system/components.rb
timriley: @solnic Would you be OK to rename this?

Issue #54, opened 2017-06-10
Move boot-related methods out of `Component`


timriley:
As noted in #44:

> I noticed a few structural things I'd like to improve later. I'm not sure if e.g. Component#bootable? and #boot_file? should exist - it feels like that's the component knowing too much about other parts of the system.

PR #65, opened 2017-11-12
Move boot related method from component

GustavoCaso:
Solves #54 

Also, I clean some files that were no using the already defined constants we have `RB_EXT` and `RB_GLOB`.


GustavoCaso:
@solnic @timriley Any thoughts?

PR #67, opened 2017-12-15
Add a plugin system for container extensions

solnic:


lib/dry/system/container.rb
```diff
@@ -103,6 +106,26 @@ def configure(&block)
...
+        # @return [self]
+        #
+        # @api private
+        def configure_logger
+          if key?(:logger)
+            self
+          elsif config.logger
+            register(:logger, config.logger)
+          else
+            config.logger = Monitor::Logger.new(config.root.join(config.log_dir).realpath.join("#{config.env}.log"))
```
timriley: @solnic May I make a suggestion for a code convention? When we're reaching _across_ dry-* gems, can we include the top-level `Dry` namespace to make it clear that we're not just using something within the current gem's codebase? Looking at things like `Monitor::Logger` in a context like this, my instinct is to think it's `Dry::System::Monitor::Logger`. Using `Dry::Monitor::Logger` would make it much clearer, IMO.

solnic: @timriley sounds good, and btw, I didn't mean to commit this here :D I'll open another PR which adds proper logging pluging :)

PR #79, opened 2018-02-08
Use namespace for component path

### LINKED ISSUE #78: Component path brokes when set a namespace that is not included in the identifier
When setting a component with a namespace that is not present in the identifier the path get mess up

```ruby
# Wrong
Dry::System::Component.new(:mailer, namespace: "mail")
#<Dry::System::Component identifier="mailer" path="mailer">

# Right
Dry::System::Component.new("mail/mailer", namespace: "mail")
#<Dry::System::Component identifier="mailer" path="mail/mailer">

# Right
 Dry::System::Component.new("mail.mailer", namespace: "mail")
#<Dry::System::Component identifier="mailer" path="mail/mailer">
```

GustavoCaso:
Fix #78 

@solnic I start using the `namespace` in the `path` that way component that specifies a namespace but is not present in the identifier will generate the correct path.

```ruby
Dry::System::Component.new(:mailer, namespace: "mail")
#<Dry::System::Component identifier="mailer" path="mail/mailer"> 
```

I removed the failing test because doesn't make much sense to specify a namespace and not have in the system. Maybe I'm missing something WDYT?

### DIFF
--- a/lib/dry/system/component.rb
+++ b/lib/dry/system/component.rb
@@ -49,11 +49,17 @@ module Dry
           ns, sep = options.values_at(:namespace, :separator)
           identifier = extract_identifier(name, ns, sep)
 
-          path = name.to_s.gsub(sep, PATH_SEPARATOR)
+          path = extract_path(identifier, ns, sep)
           loader = options.fetch(:loader, Loader).new(path)
 
           super(identifier, path, options.merge(loader: loader))
         end
+      end
+
+      def self.extract_path(identifier, ns, sep)
+        path = ns ? [ns, identifier].join(sep) : identifier
+
+        path.gsub(sep, PATH_SEPARATOR)
       end
 
       # @api private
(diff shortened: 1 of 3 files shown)


solnic:
OK so I know this is confusing, but current behavior is correct, with this fix, it actually breaks a use case where you have imported components from another container into a container with a default namespace set, let's say, to `"foo"`. Then, ie, when you try to resolve `"core.logger"`, where `"core"` is the namespace of the external component from which you imported the logger, then with this change, it would try to resolve `"foo.core.logger"`, and it breaks because there's no such thing.

It'd be good to clean this up somehow, it feels like there's too much going on in `Component` class, and it should be broken down into smaller sub-classes. I'll make sure to address this before 1.0.0.

GustavoCaso:
Thanks for the explanation @solnic. And yes it is confusing 💥

PR #82, opened 2018-03-10
Fix manual component registration with Ruby 2.5

alsemyonov:
`require` needs a `String` and doesn’t work with `Pathname` raising an exception.


solnic:
Could you add a spec covering this use case too? Given that none of the specs fail on CI under MRI 2.5.0, we clearly have missing coverage.

PR #86, opened 2018-05-25
use default value from settings Type 

### LINKED ISSUE #84: Default environment variable values aren't honored
I have a script that says, in part:

```ruby
class Application < Dry::System::Container
  use(:env)
  boot(:settings, from: :system) do
    settings do
      key :an_int, Types::Coercible::Int.default(17)
    end
  end
end
```

If I run this script and set an environment variable `AN_INT`, then `Application[:settings].an_int` gets set to the value of the variable.  If I don't set the variable then I get an error:

```
TypeError: can't convert nil into Integer
                  Integer at org/jruby/RubyKernel.java:434
                     call at org/jruby/RubyMethod.java:115
                     call at .../gems/dry-types-0.12.2/lib/dry/types/constructor.rb:47
          block in coerce at .../gems/dry-types-0.12.2/lib/dry/types/hash/schema.rb:99
(...)
```

I expect the script to work without reporting an error, and for `Application[:settings].an_int` to get its default value of 17.

https://gist.github.com/dmaze/308034415354fb0ec2e537157b3a39b8 has a more complete reproduction script.

GustavoCaso:
Fixes #84 

@solnic @andrewcroome I was able to track the issue down. Basically, we where explicitly passing the value as `nil` if the value wasn't found in `ENV` or in any of the `.env` files. To solve we just need not pass the key to the `Dry::Struct` `new` method.

So with this change instead of passing `{an_int: nil}` we pass `{}` and it will use the default value. Also this change in `Dry::Struct` helped https://github.com/dry-rb/dry-struct/pull/75

### DIFF
--- a/lib/dry/system/settings.rb
+++ b/lib/dry/system/settings.rb
@@ -41,7 +41,7 @@ module Dry
 
           attributes = schema.each_with_object({}) do |(key, type), h|
             value = ENV.fetch(key.to_s.upcase) { env_data[key.to_s.upcase] }
-            h[key] = value
+            h[key] = value if value
           end
 
           new(attributes)
--- a/spec/integration/settings_component_spec.rb
+++ b/spec/integration/settings_component_spec.rb
@@ -39,4 +39,49 @@ RSpec.describe 'Settings component' do
     expect(settings.database_url).to eql('sqlite::memory')
     expect(settings.session_secret).to eql('super-secret')
   end
+
+  context 'with default values' do
+    subject(:system) do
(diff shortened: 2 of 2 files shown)


timriley:
@GustavoCaso On the wholetThis looks right to me, thanks!

In trying to think about this from different angles, though, I did have one question: what happens if someone had _explicity_ set a "nil" or otherwise falsey value in the `env_data`? However, looking at our `Settings:FileParser`, I don't think it's possible to get a nil value out of it at all.

So I think we're good :) I'll merge this in now.

PR #87, opened 2018-05-25
raises a meaningful error if the value for a setting is incorrect

### LINKED ISSUE #85: Missing and incorrect environment variables give opaque errors
I have a sample script that says, in part,

```ruby
class Application < Dry::System::Container
  use(:env)
  boot(:settings, from: :system) do
    settings do
      key :an_int, Types::Coercible::Int.default(17)
    end
  end
end
```

If I run it like

```sh
AN_INT=20 ./dry-rb-env.rb
```

then it picks up and parses the environment variable correctly.  But if I use a non-integer value

```sh
AN_INT=foo ./dry-rb-env.rb
```

I get a very long backtrace, that doesn't mention the variable name at all:

```
ArgumentError: invalid value for Integer(): "foo"
                  Integer at org/jruby/RubyKernel.java:434
                     call at org/jruby/RubyMethod.java:115
                     call at .../gems/dry-types-0.12.2/lib/dry/types/constructor.rb:47
          block in coerce at .../gems/dry-types-0.12.2/lib/dry/types/hash/schema.rb:99
```

Similarly, if I make the setting type not have a `.default()` and don't set the environment variable, I get a

```
TypeError: can't convert nil into Integer
```

Since this is probably end-user error and not developer error, it would be nice if:

1. A missing or incorrect environment variable didn't generate a backtrace
2. These error messages included the variable name

A more complete reproduction script is at https://gist.github.com/dmaze/308034415354fb0ec2e537157b3a39b8.

GustavoCaso:
Fix #85 

@solnic @andrewcroome raise a better error if the setting value is incorrect.

### DIFF
--- a/lib/dry/system/constants.rb
+++ b/lib/dry/system/constants.rb
@@ -25,9 +25,7 @@ module Dry
       private
 
       def attributes_errors(attributes)
-        attributes.map do |key, error|
-          "#{key} #{error}"
-        end
+        attributes.map { |key, error| "#{key} #{error}" }
       end
     end
 
--- a/lib/dry/system/constants.rb
+++ b/lib/dry/system/constants.rb
@@ -11,11 +11,11 @@ module Dry
     WORD_REGEX = /\w+/.freeze
 
     DuplicatedComponentKeyError = Class.new(ArgumentError)
-    InvalidSettingValueError = Class.new(ArgumentError) do
+    InvalidSettingsError = Class.new(ArgumentError) do
       # @api private
       def initialize(attributes)
         message = <<~EOF
-          Invalid setting values for:
+          Could not initialize settings. The following settings were invalid:
 
           #{attributes_errors(attributes).join("\n")}
         EOF
@@ -25,7 +25,7 @@ module Dry
       private
 
       def attributes_errors(attributes)
-        attributes.map { |key, error| "#{key} #{error}" }
+        attributes.map { |key, error| "#{key}: #{error}" }
       end
     end
 
--- a/lib/dry/system/constants.rb
+++ b/lib/dry/system/constants.rb
@@ -11,6 +11,7 @@ module Dry
     WORD_REGEX = /\w+/.freeze
 
     DuplicatedComponentKeyError = Class.new(ArgumentError)
+    InvalidSettingValueError = Class.new(ArgumentError)
 
     # Exception raise when a plugin dependency failed to load
     #
--- a/lib/dry/system/constants.rb
+++ b/lib/dry/system/constants.rb
@@ -11,7 +11,25 @@ module Dry
     WORD_REGEX = /\w+/.freeze
 
     DuplicatedComponentKeyError = Class.new(ArgumentError)
-    InvalidSettingValueError = Class.new(ArgumentError)
+    InvalidSettingValueError = Class.new(ArgumentError) do
+      # @api private
+      def initialize(attributes)
+        message = <<~EOF
+          Invalid setting values for:
+
+          #{attributes_errors(attributes).join("\n")}
+        EOF
+        super(message)
+      end
+
+      private
+
+      def attributes_errors(attributes)
+        attributes.map do |key, error|
+          "#{key} #{error}"
+        end
+      end
+    end
 
     # Exception raise when a plugin dependency failed to load
     #
--- a/lib/dry/system/settings.rb
+++ b/lib/dry/system/settings.rb
@@ -3,6 +3,7 @@ require "dry/types"
 require "dry/struct"
 
 require "dry/system/settings/file_loader"
+require "dry/system/constants"
 
 module Dry
   module System
@@ -45,6 +46,10 @@ module Dry
           end
 
           new(attributes)
+        rescue Dry::Struct::Error => e
+          # We remove [Configuration.new] since is an internal class and could mislead users
+          part_to_avoid = (e.message.index(']')+2)...-1
+          raise InvalidSettingValueError.new(e.message[part_to_avoid])
         end
 
         def self.load_files(root, env)
--- a/lib/dry/system/settings.rb
+++ b/lib/dry/system/settings.rb
@@ -47,7 +47,7 @@ module Dry
             h[key] = value if value
           end
 
-          raise InvalidSettingValueError.new(errors) unless errors.empty?
+          raise InvalidSettingsError.new(errors) unless errors.empty?
 
           new(attributes)
         end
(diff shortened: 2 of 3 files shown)


lib/dry/system/settings.rb
```diff
@@ -45,6 +46,10 @@ def self.load(root, env)
           end
 
           new(attributes)
+        rescue Dry::Struct::Error => e
+          # We remove [Configuration.new] since is an internal class and could mislead users
+          part_to_avoid = (e.message.index(']')+2)...-1
```
davydovanton: do you have a full string with error? Maybe we can create a easier way to trim it 🤔

timriley: Yeah, this does feel like a brittle way to get a nicer error message out.

For reference, here's the full `e.message` in this context:

```
[Configuration.new] "foo" (String) has invalid type for :integer_value violates constraints (type?(Integer, "foo") failed)
```

Since we own dry-struct, perhaps we can update `Dry::Struct::Error` to carry some nicer data for us to use?

I would imagine that we could add some separate attributes there for the key name and type failure message?

_Finally,_ if we're looking at better handling of these settings type errors, there's one other thing I'd like us to do that I think will make this experience a lot nicer: try applying _all_ the settings and collecting _all_ errors that occur as a part of this process, so if there's e.g. 3 missing or invalid values in someone's config, they don't have to check 3 times to fix them all up. Instead, they should see a full list of all the failing settings.

I think this would involve doing something a little different than just a straight-up `new(attributes)`, e.g. actually checking each key/val against the Configuration class' schema, but I think it'd make for a **dramatically** improved experience, which I think is an important thing for something like a dry-system application's settings.

@GustavoCaso @davydovanton What do you think?

@GustavoCaso Would you be interested in giving this a go? I figure now's a good time if you already have your head in this part of the code.

GustavoCaso: I'll see if we can improve `Dry::Struct::Error` to see if we can provide better data for use.

And the last suggestion I think is great and would not require a huge code change in order to apply it. 

Thanks for comments @timriley and @davydovanton

GustavoCaso:
@timriley @davydovanton there is no need to modify `Dry::Struct` we can use `Dry::Types#try` method to check if the actual value is valid against the setting schema, here is the code that does that https://github.com/dry-rb/dry-system/pull/87/commits/6b811ff2c1dc8103768e6284adb5ac31e88179e8#diff-28b9ac729d521a10a9c47a534da39e71R61 what I can see from this approach is that errors from `Dry::Types` do not always follow the same structure https://github.com/dry-rb/dry-system/pull/87/commits/6b811ff2c1dc8103768e6284adb5ac31e88179e8#diff-5aa61a56c8dbc0f087b9c14b6d818b10R84 maybe we can improve the error message on `Dry::Types` or `Dry::Logic` WDYT?

lib/dry/system/constants.rb
```diff
@@ -11,6 +11,25 @@ module System
...
+          #{attributes_errors(attributes).join("\n")}
+        EOF
+        super(message)
+      end
+
+      private
+
+      def attributes_errors(attributes)
+        attributes.map do |key, error|
+          "#{key} #{error}"
```
davydovanton: style question: maybe split it to one line?

lib/dry/system/settings.rb
```diff
@@ -38,19 +39,30 @@ def self.setting(*args)
 
         def self.load(root, env)
           env_data = load_files(root, env)
+          errors = {}
 
           attributes = schema.each_with_object({}) do |(key, type), h|
             value = ENV.fetch(key.to_s.upcase) { env_data[key.to_s.upcase] }
+            check_for_error(key, type, value, errors)
```
timriley: The way `check_for_error` works here makes me feel a bit uncomfortable: it's inside a loop mutating `h` to provide a meaningful return value, while along the way you're passing a _different_ object to to be mutated by some other method far away. Just doesn't clean to me.

I could see us cleaning it up in a couple of ways:

1. Initializing `attributes = {}` above and turning this into an `each` loop, then assigning to `attributes` and `errors` (if required) for each setting. This is just being honest to ourselves that this particular code is procedural :)
2. Assigning `attributes` as we do currently (with the `check_for_error` line removed), then adding _another_ `each_with_index` afterwards to build up an errors hash. This would be a bit slower though because we're looping through the attributes twice.

(1) is probably fine here, I think. What do you reckon?

GustavoCaso: Tim totally agree with you, it doesn't look clean. 

Between both suggestions, I would go for the first option, fewer iterations.

GustavoCaso:
@timriley applied all comments. It actually looks much more clean and understandable

spec/integration/settings_component_spec.rb
```diff
@@ -73,16 +73,16 @@
...
-        Dry::System::InvalidSettingValueError,
+        Dry::System::InvalidSettingsError,
         <<~EOF
-          Invalid setting values for:
+          Could not initialize settings. The following settings were invalid:
 
-          integer_value type?(Integer, "foo")
-          coercible_value invalid value for Integer(): "foo"
+          integer_value: type?(Integer, "foo")
+          coercible_value: invalid value for Integer(): "foo"
```
timriley: Next job for us (definitely outside of the scope of this PR) is to make dry-types error messages less cryptic than `type?(Integer, "foo")` 😅

GustavoCaso: Couldn't agree more with you 😄

lib/dry/system/settings.rb
```diff
@@ -40,11 +40,14 @@ def self.setting(*args)
...
           env_data = load_files(root, env)
           errors = {}
+          attributes = {}
 
-          attributes = schema.each_with_object({}) do |(key, type), h|
+          schema.each do |key, type|
             value = ENV.fetch(key.to_s.upcase) { env_data[key.to_s.upcase] }
-            check_for_error(key, type, value, errors)
-            h[key] = value if value
+            result = check_value_against_type?(type, value)
```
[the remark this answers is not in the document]
GustavoCaso: True is quite generic. I did call it `result` because actually it returns a `Result` but not everyone knows that. `type_check` is fine with me.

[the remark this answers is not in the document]
GustavoCaso: We can even remove the method, since is just one line. WDYT?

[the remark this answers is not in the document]
timriley: Yeah, having a one-liner right in place makes sense.

PR #89, opened 2018-05-31
Add inflector to container settings

flash-gordon:
This makes it easier to provide a custom inflector. At the same time, this
change is not backward compatible with existing custom loaders since now their
constructors have to accept 2 arguments.

image

It's my first PR here, please be gentle!

### DIFF
--- a/lib/dry/system/component.rb
+++ b/lib/dry/system/component.rb
@@ -1,6 +1,7 @@
 require 'concurrent/map'
 
 require 'dry-equalizer'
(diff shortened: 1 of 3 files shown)


flash-gordon:
@GustavoCaso no, it wouldn't because it's the problem on the calling side which now passes two args. I brought back the default arg, though, on morning I decided it works better :)

GustavoCaso:
@flash-gordon awesome!

PR #91, opened 2018-06-16
add Container#stop

### LINKED ISSUE #90: How `stop` callback is invoked?
Consider following example from documentation

```
Application.finalize(:persistence) do |container|
  start do
    require 'sequel'
    container.register('persistence.db', Sequel.connect(ENV['DB_URL']))
  end

  stop do
    db.disconnect
  end
end
```

When `stop` callback (which calls `db.disconnect`) will be called? Is there a method to *stop* all components at once (like `finalize!` which *starts* all components)

This case is not covered in documentation...

GustavoCaso:
Fixes #90 

@solnic @mainameiz I have added the `stop` method in `Container`.

It also raises an error when trying to stop a component that has not started yet.

### DIFF
--- a/lib/dry/system/errors.rb
+++ b/lib/dry/system/errors.rb
@@ -62,9 +62,9 @@ module Dry
     # Error raised when trying to stop a component that hasn't started yet
     #
     # @api public
-    ComponentNotStaredError = Class.new(StandardError) do
-      def initialize(companent_name)
-        super("You can not stop a component that has not been started yet. Please use `start` method to start the component. system.start[:#{companent_name}]")
+    ComponentNotStartedError = Class.new(StandardError) do
+      def initialize(component_name)
+        super("You can not stop a component +#{component_name}+ has not been started.")
       end
     end
   end
--- a/lib/dry/system/errors.rb
+++ b/lib/dry/system/errors.rb
@@ -64,7 +64,7 @@ module Dry
     # @api public
     ComponentNotStartedError = Class.new(StandardError) do
       def initialize(component_name)
-        super("You can not stop a component +#{component_name}+ has not been started.")
+        super("component +#{component_name}+ has not been started")
       end
     end
   end
--- a/lib/dry/system/errors.rb
+++ b/lib/dry/system/errors.rb
@@ -58,5 +58,14 @@ module Dry
         super("component identifier #{name.inspect} must be a symbol")
       end
     end
+
+    # Error raised when trying to stop a component that hasn't started yet
+    #
+    # @api public
+    ComponentNotStaredError = Class.new(StandardError) do
+      def initialize(companent_name)
+        super("You can not stop a component that has not been started yet. Please use `start` method to start the component. system.start[:#{companent_name}]")
+      end
+    end
   end
 end
--- a/spec/integration/boot_spec.rb
+++ b/spec/integration/boot_spec.rb
@@ -63,11 +63,62 @@ RSpec.describe Dry::System::Container, '.boot' do
         start do
           register('db.conn', Test::Db.new(established?: true))
         end
-      end
 
+        stop do
+          db.conn.established = false
+        end
+      end
       conn = system['db.conn']
 
       expect(conn).to be_established
+    end
+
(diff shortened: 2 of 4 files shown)


timriley:
@GustavoCaso Rather than raise an error, couldn't this just gracefully skip any components not yet booted?

lib/dry/system/errors.rb
```diff
@@ -58,5 +58,14 @@ def initialize(name)
         super("component identifier #{name.inspect} must be a symbol")
       end
     end
+
+    # Error raised when trying to stop a component that hasn't started yet
+    #
+    # @api public
+    ComponentNotStaredError = Class.new(StandardError) do
```
timriley: There's a typo here. It should be `ComponentNotStaredError`.

lib/dry/system/errors.rb
```diff
@@ -58,5 +58,14 @@ def initialize(name)
...
+      def initialize(companent_name)
+        super("You can not stop a component that has not been started yet. Please use `start` method to start the component. system.start[:#{companent_name}]")
```
timriley: I feel like we could tighten this up a bit:

```ruby
"component +#{component_name}+ has not been started"
```

(I'm not sure we need the usage tip at this point)

lib/dry/system/errors.rb
solnic: `ComponentNotStaredError` => `ComponentNotStartedError`

lib/dry/system/errors.rb
solnic: `companent_name` => `component_name`

mainameiz:
@GustavoCaso Thanks a lot!

lib/dry/system/errors.rb
GustavoCaso: Bloody typo 🤣

lib/dry/system/errors.rb
GustavoCaso: @timriley we can remove the tip, and I will fix the typo as well

GustavoCaso:
@timriley @solnic I have fixed the typos and remove the tip from the error message. If everything looks good to you. I can merge and update the CHANGELOG and start working on the `shutdown` feature 😄

lib/dry/system/errors.rb
```diff
@@ -58,5 +58,14 @@ def initialize(name)
         super("component identifier #{name.inspect} must be a symbol")
       end
     end
+
+    # Error raised when trying to stop a component that hasn't started yet
+    #
+    # @api public
+    ComponentNotStartedError = Class.new(StandardError) do
+      def initialize(component_name)
+        super("You can not stop a component +#{component_name}+ has not been started.")
```
timriley: @GustavoCaso Would you mind changing it to `"component +#{component_name}+ has not been started"`?

spec/integration/boot_spec.rb
```diff
@@ -63,11 +63,62 @@ class Db < OpenStruct
...
+      end
+      system.start(:db)
+
+      conn = system['db.conn']
+      system.stop(:db)
+
+      expect(conn.established).to eq false
+    end
+
+    it 'raises an error when trying to stop a component that has not bee started' do
```
timriley: typo, `bee` -> `been`

spec/integration/boot_spec.rb
```diff
@@ -63,11 +63,62 @@ class Db < OpenStruct
...
+        stop do
+          db.conn.established = false
+        end
+      end
       conn = system['db.conn']
 
       expect(conn).to be_established
     end
+
+    it 'allows stop component' do
```
timriley: This would read better as "allows component to be stopped" :)

timriley:
@GustavoCaso Left a couple more wording notes for you, happy for you to merge after that 👍

lib/dry/system/errors.rb
GustavoCaso: Sure thing. Sorry if I did not removed in the last commit, I understood only removing the tip part

GustavoCaso: @timriley done

Issue #94, opened 2018-07-19
Support conditional booting of providers


davydovanton:
I found the interesting case for improvement.

In my hanami app I have server and rake "instances". For rake I need to use `aws` dependency but in server booting, I don't need it. In this case, it will be better for me to load `aws` only for rake env.

In hanami I can use `Hanami.app?(:rake)` ([docs](http://hanamirb.org/guides/1.2/projects/selectively-boot-apps/)). And it looks like this:

```ruby
if Hanami.app?(:rake)
  Container.boot :rake_test do |container|
    start do
      container.register :rake_test, Object.new
    end
  end
end
```

Also, I updated rakefile:

```ruby
ENV['HANAMI_APPS'] = 'rake'
```

----------------------------

So, what I suggest. Create a optional argument for booting dependency, something like this:

```ruby
Container.boot(:rake_test, condition: -> { Hanami.app?(:rake) }) do |container|
  start do
    container.register :rake_test, Object.new
  end
end
```

And in this case, dry-system will boot dependency only with `true` in condition. Also, it can be related to https://github.com/dry-rb/dry-system/issues/73 but I'm not sure. 

WDYT?

cc @solnic @timriley @GustavoCaso

timriley:
@davydovanton I like this idea. 👍 

And your suggested implementation is generic enough such that it would also e.g. Hanami to come in and provide a more framework-specific API to wrap it and set that `condition:` proc as required.

(For similar scenarios in our apps we've designed our rake tasks so that they never fully boot the container, instead only booting specific components as required. This works fine if all those bootable components are _also_ needed when fully booting the container (i.e. to run the web app), but it wouldn't fit in the situation you describe, where you need a bootable component for specific tasks that should _not_ always boot when the full container boots.)

My only other thought here is (of course) about naming. I wonder whether `when:` or `boot_if:` or something might be a better name for the option.

GustavoCaso:
@davydovanton I also like the idea 🎉 

 As @timriley mentioned is generic enough.

I do not have an example as to how we do it in our apps 😢.

And regarding naming, I think it reads really well:

```ruby
Container.boot(:rake_test, when: -> { Hanami.app?(:rake) }) do |container|
  start do
    container.register :rake_test, Object.new
  end
end
```

I'm not sure if `when` being a ruby keyword would became a problem ❓

jodosha:
[...]

PR #95, opened 2018-07-26
Allow to use conditions for booting components

davydovanton:
Reference to https://github.com/dry-rb/dry-system/issues/94

### DIFF
--- a/lib/dry/system/container.rb
+++ b/lib/dry/system/container.rb
@@ -202,7 +202,7 @@ module Dry
         #   # system/boot/db.rb
         #   #
         #   # Optional component registration
-        #   MyApp.boot(:db, when: -> { !ENV['DB_URL'].empty? }) do |container|
+        #   MyApp.boot(:db, when_condition: -> { !ENV['DB_URL'].empty? }) do |container|
         #     start do
         #       require 'db'
         #       DB.configure(ENV['DB_URL'], logger: logger)
@@ -227,12 +227,13 @@ module Dry
         # @return [self]
         #
         # @api public
-        def boot(name, opts = {}, &block)
-          boot_condition = opts.delete(:when)
-
-          if boot_condition
-            raise(InvalidBootConditionTypeError, name) unless boot_condition.respond_to?(:call)
-            return unless boot_condition.call
+        def boot(name, when_condition: nil, **opts, &block)
+          if when_condition
+            begin
+              return unless when_condition.call
+            rescue NoMethodError => e
+              e.name == :call ? raise(InvalidBootConditionError, name) : raise
+            end
           end
 
           if components.key?(name)
--- a/lib/dry/system/container.rb
+++ b/lib/dry/system/container.rb
@@ -201,8 +201,25 @@ module Dry
         #
         #   # system/boot/db.rb
         #   #
-        #   # Optional component registration
-        #   MyApp.boot(:db, when_condition: -> { !ENV['DB_URL'].empty? }) do |container|
+        #   # Optional component registration based on condition broc
+        #   MyApp.boot(:db, if: -> { !ENV['DB_URL'].empty? }) do |container|
+        #     start do
+        #       require 'db'
+        #       DB.configure(ENV['DB_URL'], logger: logger)
+        #       container.register(:db, DB.new)
+        #     end
+        #   end
+        #
+        #   # system/boot/db.rb
+        #   #
+        #   # Optional component registration based on container class method
+        #   class MyApp
+        #     def self.allow_boot_db?
+        #       false
+        #     end
+        #   end
+        #
+        #   MyApp.boot(:db, if: :allow_boot_db?) do |container|
         #     start do
         #       require 'db'
         #       DB.configure(ENV['DB_URL'], logger: logger)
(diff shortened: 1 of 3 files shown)


lib/dry/system/container.rb
```diff
@@ -217,6 +228,13 @@ def import(other)
         #
         # @api public
         def boot(name, opts = {}, &block)
+          boot_condition = opts.delete(:when)
+
+          if boot_condition
+            raise(InvalidBootConditionTypeError, name) unless boot_condition.respond_to?(:call)
+            return unless boot_condition.call
+          end
```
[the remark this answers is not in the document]
GustavoCaso: Much cleaner that way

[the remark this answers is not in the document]
radanskoric: If `boot_condition` has a bug and throws a `NoMethodError` somewhere inside its code, it's going to be masked by this rescue making it harder to debug for the developer. It should at least do:
```
begin
  return unless boot_condition.call
rescue NoMethodError => e
  e.name == :call ? raise(InvalidBootConditionTypeError, name) : raise
end
```

It'll still mask the error if code inside `boot_condition` has a bug where it invokes `call` on an object that doesn't have that method but that's maybe an acceptable compromise.

davydovanton:
sorry folks, I forgot about this PR 😢

Drop mutation and rename option to `when_condition` instead `when`

solnic:
> Drop mutation and rename option to when_condition instead when

I'm thinking...why not `if`? so:

``` ruby
MyApp::Container.boot(:foo, if: :bar?) { ... }
```

davydovanton:
@solnic 2 points:

1. it will  be harder to work with any default ruby keywords;
2. we put a block into the option. in case of `if` I expect `Bool` object instead `proc`. Not sure that it's a blocker but I think it can be confused. WDYT?

solnic:
> it will be harder to work with any default ruby keywords

Why?

> we put a block into the option. in case of if I expect Bool object instead proc. Not sure that it's a blocker but I think it can be confused. WDYT?

My understanding is that this is an option that will be used to evaluate provided proc so it shouldn't matter how it's called. I would also prefer if it accepted a container method name, so that you can implement these checks in the container class. This way you could ie stub it in tests and it would also make defining bootable components cleaner.

davydovanton:
> My understanding is that this is an option that will be used to evaluate provided proc so it shouldn't matter how it's called

not sure that I understand your right. Could you share some code examples for better understanding?

lib/dry/system/container.rb
```diff
@@ -216,7 +227,15 @@ def import(other)
         # @return [self]
         #
         # @api public
-        def boot(name, opts = {}, &block)
+        def boot(name, when_condition: nil, **opts, &block)
+          if when_condition
+            begin
+              return unless when_condition.call
```
solnic: @davydovanton here, instead of `when_condition.call`, we'd do:

``` ruby
case when_condition
  when Symbol then __send__(when_condition)
  when Proc then when_condition.call
  else
     raise "OOOOPS"
  end
```

davydovanton: got it, thanks!

davydovanton:
@solnic I updated PR. Rename option to `if` and add ability to use container class method name as a `if` condition. Could you check PR when you'll have time?

lib/dry/system/container.rb
```diff
@@ -216,7 +244,22 @@ def import(other)
...
+
+          if boot_condition
+            begin
+              case boot_condition
+              when Symbol
+                return unless __send__(boot_condition)
+              else
+                return unless boot_condition.call
+              end
+            rescue NoMethodError => e
```
timriley: Are you intentionally wrapping both cases (boot_condition being either a Symbol or a proc) in this begin/rescue? Wouldn't it be better just to wrap the `boot_condition.call` case?

davydovanton: hey, the general idea here: user can use bool value:

```ruby
MyApp.boot(:db, if: false) do |container|
  # ...
end
```

In this case, the user gets `NoMethodError ` and it will be bad from UX perspective. Instead of this, I suggest to raise a normal exception and say that a user can't use this object as an argument. WDYT?

timriley: @davydovanton Yeah, I think this'd be the better way to go

timriley: @davydovanton I wonder if we can actually expand the range of values we accept a little...

Firstly, given a `String` or `Symbol` object — make it so _both_ of these are considered container identifiers, and we resolve the matching object from the container accordingly.

Then, by this point, we should have either a:

(1) Callable object, or
(2) Something that's otherwise truthy or falsy, per standard Ruby behaviour

So by this point here we just .call() if we can, otherwise rely on the truthy/falsy result of whatever expression we've been passed.

How does that feel?

timriley:
@davydovanton Left you another comment. After we address that, I think we can merge this thing!

Morozzzko:
I know there's thin hope, but what's up with the PR?

If it just needs to be rebased and updated, feel free to hit me up and I'll do that. I felt like this feature would be awesome to have

timriley:
@Morozzzko I'll be taking this on as part of preparing for 1.0 over the next month.

In the meantime, I'd love to hear more examples of real-world uses for this feature, just to make sure I can cover as many needs as possible. Would you mind sharing how you'd want to use it?

PR #97, opened 2018-12-01
Testing against new dry-container

flash-gordon:


solnic:
@flash-gordon great!

flash-gordon:
Overall, the changes are minor, and I think we can have it merged. Releases of dry-container and dry-system have to be coordinated

PR #98, opened 2018-12-16
Add Container#shutdown!

### LINKED ISSUE #92: Add Container#shutdown
As mentioned in the issue https://github.com/dry-rb/dry-system/issues/90

The lifecycle process is missing the stop method. It has been addressed in the PR https://github.com/dry-rb/dry-system/pull/91

@solnic mentioned also It would be interesting to have a `shutdown` method that goes through all booted components and stop them.

GustavoCaso:
Fix #92 

@davydovanton @solnic I added the `shutdown!` method to stop all components.

I thought that if a component is not started, we shouldn't raise `ComponentNotStartedError` we simply skip it. My reason behind it is whatever the reason the user of `dry-system` for calling `shutdown!` is to stop everything, so if one component has not been stated it shouldn't affect, actually less work for us :)

### DIFF
--- a/lib/dry/system/booter.rb
+++ b/lib/dry/system/booter.rb
@@ -70,6 +70,15 @@ module Dry
       end
 
       # @api private
+      def shutdown
+        components.each do |component|
+          next unless booted.include?(component)
+
+          stop(component)
+        end
+      end
+
+      # @api private
       def init(name_or_component)
         with_component(name_or_component) do |component|
           call(component) do
@@ -100,7 +109,10 @@ module Dry
       def stop(name_or_component)
         call(name_or_component) do |component|
           raise ComponentNotStartedError.new(name_or_component) unless booted.include?(component)
+
           component.stop
+          booted.delete(component)
+
           yield if block_given?
         end
       end
--- a/lib/dry/system/container.rb
+++ b/lib/dry/system/container.rb
@@ -358,6 +358,11 @@ module Dry
           self
         end
 
+        def shutdown!
+          booter.shutdown
+          self
+        end
+
         # Sets load paths relative to the container's root dir
         #
         # @example
--- a/spec/integration/boot_spec.rb
+++ b/spec/integration/boot_spec.rb
@@ -2,6 +2,43 @@ require 'ostruct'
 
 RSpec.describe Dry::System::Container, '.boot' do
   subject(:system) { Test::Container }
+  let(:setup_db) do
+    system.boot(:db) do
+      init do
+        module Test
+          class Db < OpenStruct
+          end
+        end
+      end
+
+      start do
+        register('db.conn', Test::Db.new(established: true))
+      end
+
+      stop do |container|
+        container['db.conn'].established = false
+      end
+    end
+  end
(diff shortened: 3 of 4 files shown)


solnic:
CI failed

lib/dry/system/booter.rb
```diff
@@ -69,6 +69,15 @@ def finalize!
         freeze
       end
 
+      # @api private
+      def shutdown
+        components.each do |component|
+          stop(component)
+        rescue ComponentNotStartedError
```
v-kolesnikov: We could just iterate over `booted` and don't have to handle `ComponentNotStartedError` exception. Whay not?

v-kolesnikov: I also find it strange, that booter.stop(<component>) doesn't upload a component from `booted` registry, so it doesn't allow to start the same component twice. Or it is by design? If yes, why? @solnic @GustavoCaso

timriley: I doubt this is by design, I think it's an oversight and something we should fix.

solnic: It should be allowed to start something multiple times, so it's just not finished, that's all.

flash-gordon:
CI failed because of using `rescue` on the block level, it's 2.5+ syntax. Fancy stuff doesn't become available that fast if you support more than one version of the language :)

lib/dry/system/booter.rb
GustavoCaso: @v-kolesnikov nice catch about using `booted` instead of components. It would also fix CI since we wouldn't need the `rescue` part @flash-gordon pointed out ❤️

GustavoCaso:
@v-kolesnikov At the end, I had to use `components` instead of `booted` because after adding the code for deleting the component from the `booted` array https://github.com/dry-rb/dry-system/pull/98/files#diff-9300e9bdf4fe77c133ba34f93e8498bfR116 the specs were failing. My two cents is that we are iterating over `booted` and modifying it at the same time by deleting elements from it, is not a good thing to do

GustavoCaso:
@solnic @timriley @flash-gordon I have updated the code to make CI pass as well to remove a component from `booted` when stopping them.

lib/dry/system/booter.rb
```diff
@@ -69,6 +69,17 @@ def finalize!
...
+      # @api private
+      def shutdown
+        components.each do |component|
+          begin
+            stop(component)
+          rescue ComponentNotStartedError
+            next
+          end
+        end
+      end
```
solnic: IIRC there's a way to check which components have been started. Maybe you could use that instead of rescuing from `ComponentNotStartedError`?

GustavoCaso: @solnic Yeah makes much more sense. Looks like I have fixated mind about the rescue block. Thank you for the patients.

Done https://github.com/dry-rb/dry-system/pull/98/commits/bec81caf8d1d25051f659ab461ee8a72f894ae99

solnic:
LGTM but please rebase/squash before merging

PR #99, opened 2018-12-27
Add the ability to set the memoization option to the default `AutoRegistrar`

arielvalentin:
Per this (conversation)[https://discourse.dry-rb.org/t/lazy-instantiation-and-memoization-with-container-auto-registration/79/11], this PR introduces the ability for users to configure memoization of auto-registered objects, e.g.
```
  configure do |config|
    config.auto_registration.memoize = true
  end
  load_paths!('lib', 'system')
  auto_register!('lib')
end
``` 
- This change preserves the existing default behavior and will require users to explicitly change the setting if they want to enable memoization
- Custom implementations of `AutoRegistrar` will need to change in order to use this new setting


GustavoCaso:
Hi @arielvalentin 
Thank you so much for putting this together ❤️ 

I'm wondering if some user might want `memoize` some folder and some other folders not. With the current solution that would be possible.

Maybe we could make it that the configuration per `auto_register!` block

```ruby
  load_paths!('lib', 'system')
  auto_register!('lib') { |auto_register_config| auto_register_config.memoize = true }
end
```

This way the users have more control over which components are `memoize` and we do not add more settings to the container class.

If we decide to go in this direction the configuration that is yield is here https://github.com/dry-rb/dry-system/blob/master/lib/dry/system/auto_registrar/configuration.rb

arielvalentin:
@GustavoCaso Thanks so much for taking the time to review this PR! That sounds like an great idea. 

I did head down that route first, i.e. adding the property to the `Configuration` class, however I had a question about that class defining it's own `setting` DSL as opposed to using `dry-configurable`. It it still preferred to use the custom setting method with a default value initialized in the constructor?

arielvalentin:
@GustavoCaso I have pushed a small change that uses an `attr_accessor` to achieve the API you are looking for. Is that acceptable or are you looking more for a DSL based approach that is symmetric with `dry-configurable`?

GustavoCaso:
Hi @arielvalentin 

Thank you so much for taking the time to review and update the PR. The code LGTM.

I left a small comment.

examples/custom_configuration_auto_register/system/container.rb
```diff
@@ -1,6 +1,10 @@
 require 'dry/system/container'
 
 class App < Dry::System::Container
+  configure do |config|
+    config.auto_registration.memoize = true
```
GustavoCaso: We should probably update this example too 😄

GustavoCaso:
Regarding the DSL like `dry-configurable` I think we could keep the code as it is since we do not know how many more configurations are going to be added, and if in the future the number of `settings` grow too much and we new to update we could do it. WDYT?

examples/custom_configuration_auto_register/system/container.rb
arielvalentin: Oops! I totally forgot about this!

arielvalentin:
> Regarding the DSL like `dry-configurable` I think we could keep the code as it is since we do not know how many more configurations are going to be added, and if in the future the number of `settings` grow too much and we new to update we could do it. WDYT?

@GustavoCaso My only concern would be it being asymmetric from other properties, but then again the other settings are procs and do not have a similar usage to that of a primitive value. As long as you are 👍  with it, then I am 👍

arielvalentin:
@GustavoCaso I have pushed up another change. Would you like for me to squash the commits?

GustavoCaso:
@timriley I mentioned you, since you said in the [conversation](https://discourse.dry-rb.org/t/lazy-instantiation-and-memoization-with-container-auto-registration/79/11) you would like to see the PR. From my side LGTM

lib/dry/system/auto_registrar/configuration.rb
```diff
@@ -12,11 +12,11 @@ class AutoRegistrar
       class Configuration
         DEFAULT_INSTANCE = -> component { component.instance }.freeze
         FALSE_PROC = -> * { false }.freeze
+        FALSE_DEFAULT = false
```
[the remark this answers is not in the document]
arielvalentin: I only did that for symmetry. I will inline the constant.

lib/dry/system/auto_registrar/configuration.rb
```diff
@@ -28,10 +28,13 @@ def self.setting(name)
         setting :exclude
         setting :instance
 
+        attr_accessor :memoize
+
```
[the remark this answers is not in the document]
GustavoCaso: @solnic I don't remember a real reason why actually. We can use `dry-configurable`.

Would you like it to be done in this PR?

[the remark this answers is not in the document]
solnic: > @GustavoCaso Would you like it to be done in this PR?

No. I was just curious because I can't remember either.

> @arielvalentin Does dry-configurable support the desired syntax? 

It does. We can consider using it instead but that's for another PR.

spec/unit/container/auto_register_spec.rb
```diff
@@ -51,6 +51,50 @@ class Test::Container < Dry::System::Container
...
+
+    context 'with explicit overrides' do
+      context 'with memoization enabled' do
+        it "memoizes results" do
+          container.auto_register!('components') do |config|
+            config.memoize = true
+          end
+
+          expect(container['foo']).to be_an_instance_of(Foo)
+          expect(container['foo']).to eql(container['foo'])
```
solnic: This should be

``` ruby
expect(container['foo']).to be(container['foo'])
```

Otherwise it only checks equality.

spec/unit/container/auto_register_spec.rb
```diff
@@ -51,6 +51,50 @@ class Test::Container < Dry::System::Container
...
+        end
+      end
+
+      context 'with memoization disabled' do
+        it "does not memoize results" do
+          container.auto_register!('components') do |config|
+            config.memoize = false
+          end
+          expect(container['foo']).to be_an_instance_of(Foo)
+          expect(container['foo']).to_not eql(container['foo'])
```
solnic: In a similar fashion as above, this should use `to_not be`.

arielvalentin:
Awesome!!! 

Is there documentation I can review that describes the release process and current roadmap?

PR #101, opened 2019-01-04
Use class attribute for container namespace_separator setting

timriley:
This is necessary to work with the latest dry-container, which (for dry-rb-wide dependency graph reasons) has switched from dry-configurable to simple dry-core class attributes for its config.

### DIFF
--- a/lib/dry/system/importer.rb
+++ b/lib/dry/system/importer.rb
@@ -18,7 +18,7 @@ module Dry
       # @api private
       def initialize(container)
         @container = container
-        @separator = container.config.namespace_separator
+        @separator = container.namespace_separator
         @registry = {}
       end
 


GustavoCaso:
@timriley nice catch 🎉 

I'm wondering if we should add some specs for `Importer`?

solnic:
How come this doesn't make any test crash?

GustavoCaso:
@timriley since the change in `dry-container` was done by me I don't mind adding some test for the `Importer`.

If you don't mind I will add them in this PR.

timriley:
@GustavoCaso I'd hold off on doing any extra work in this branch. @flash-gordon is going to revert dry-container back to using dry-configurable again (given we're looking at adjusting the dry-configurable API and making a dry-types/dry-struct dependency optional).

PR #106, opened 2019-02-22
Endash tweaks

endash:
⚠️ In case it isn't clear, this ticket is for bookmarking, documentation, and discussion. I anticipate cutting it up into multiple PRs for merge. ⚠️ 

## Big picture overview

The big thing is that plugins have been refactored to resemble adapter classes, and `Booter`, `AutoRegistrar`, et al each have a plugin that manages their integration with `Container`. The `Booter`, `AutoRegistrar`, `Importer`, and `ManualRegistrar` core classes have been refactored mostly to tighten up their encapsulation, a side effect of which is to make them container-agnostic (they all work with vanilla `Dry::Container`s.) Also there have been some nomenclature changes to avoid ambiguity around terms like "component," and some minor behavioral tweaks.

## Nomenclature

There is currently an overlap in how terms are used, particularly "component." In general it seems to be meant as "an object registered on the container," but in the context of booting it usually, but not always, means an instance of `Bootable` registered on the `Booter` via `Container#boot`, which can register 0, 1, or more components, the names of which are not necessarily the same as the `Bootable`'s identifier (i.e., there is no 1:1 correspondence between components and _components_. Read that second "components" with a goofily bad fake accent, like football and _futbol_.)

It can also mean "an object that _will be_ registered on the container but has not yet been." More abstractly, it could mean "an object of some significance on which your app depends." It's confusing enough on its own, but I think there are also some subtle design consequences throughout the codebase, particularly in the interplay between `Component` and `Bootable` outside of the `Booter`.

I've cut through the Gordian knot by simply giving more specific terms for each use, or by eliminating the equivocation directly, via shifting responsibilities for the objects in question. The final result is that "component" only has its general, abstract meaning, usually just "some part of your app." _There is no first-class "component" or `Component` entity._

## Architecture

This has shifted from a mini-monolith (minilith? has anyone coined that yet?) to fully modular and plugin-based. There is already plugin support, of course, but I've made them first-class entities and shifted them to the core of the architecture.

### Common, low-level classes

There are a few important low-level common classes. `Provider` and `Identified` (previously `Bootable` and `Component`, respectively) are more-or-less "dumb" configuration objects, and `Loader` now handles both requiring files and loading objects. `Provider` _provides_, via `register`, 0 or more components, and just stores the `start`, `stop`, `init`, hooks, config, and the rest of its definition as ivars. It is consumed by the `Booter` and is pretty much immediately wrapped in a `Lifecycle`, a combined FSM and provider DSL execution-context, that handles the mechanics of the boot process for an individual provider.

`Identified` reifies the notion of "some identifier yet to be registered on the container." It is just a decorator for an identifier, basically, with methods for turning it into a path, checking for existence, and applying/checking a namespace. `Identified`s are created/consumed by autoregistration and by the `Container` pre-finalization, and are passed to plugins that have a `#missing_key` method.

### Base

The core class is `Base`. `Base` is just `Dry::Container` plus really basic event hooks/triggers, a `PluginsManager`, and overridden `configure`/`finalize` methods to tie them together. By itself, it's a blank canvas, and incrementally adds functionality to `Dry::Container`.

### Booter, AutoRegistrar, et al.

The `Booter`, `AutoRegistrar`, and `Importer` classes are where most of the magic happens. They've been promoted a rank and have had their encapsulation tightened up. Their core classes depend only on receiving an object with `merge` and `register` methods (i.e., a `Dry::Container`), plus a modicum of configuration they get provided directly. 

### Plugins

Plugins are first-class entities: they're long-lived objects that are initialized with access to the container, can have events triggered on them (`after_configure`, `finalize`, etc), and are responsible for providing any instance(s) of collaborators. They're adapters, essentially, mediating between the container on one hand and the `Booter` et al. on the other hand. 

Plugins are installed directly into a container that descends from `Base`, via `use`. The global registry for handling plugin dependencies has been factored out, since the main benefit of such a scheme (implementation independence, dependency inversion, what-have-you) is arguably unrealizable without first-class interfaces, and everything ends up boiling down to a constant in the code anyway. Developers still ultimately need to be aware of a plugin's required registry identifiers and the interface that prerequisite is expected to meet, and then ensure that there is an appropriate plugin also registered, exacting a complexity cost even in the trivial situation where the dependency is built-in, like `env`. Better just to be aware that you have to use `env`, too, if you want to use `foo`. 

I'm of a mind that the best design is to put the onus on developers working with plugins to simply be aware of the antecedents and consequences of a plugin. This is by way of saying that plugins should be simple, comprehensible, and easy to reason about, firstly to avoid hard-to-resolve bugs, and secondly to facilitate understanding of how the entire composed system functions. If a plugin requires the `env` plugin, then that's something the user should know about, understand, and take ownership of. Complex plugin graphs are to be avoided.

The flipside of that is when a complex plugin simply _composes_ other plugins, as a convenience, in which case it can just install those sub-plugins itself. Again, though, this should remain simple enough that the plugin could be de-composed if needed. Composition, as opposed to a dependency, is a situation where indirection like a registry makes some sense, so that the composed plugin could be selectively overriden. IMO, though, it's better to just encourage simple designs and implementations, at the cost of a tiny bit more configuration code in a small minority of use cases. 

As a side benefit, a plugin is now mostly just regular old ruby code, with any dependencies required at the top of the file, and a (mostly) PORO instance for encapsulating state and segregating the interface.

### Bootables/Providers

The single biggest change in behavior is that `Provider`s' (née `Bootable`) identifiers aren't assumed, at a low level, to correspond to keys on the `Container`. Because there isn't any guarantee or even necessarily any reason to believe that a `Provider`'s identifier corresponds to a resolvable key, or vice versa, the `Booter` plugin is mostly treated as its own separate system.

However, all that said: In order to preserve the desired functionality of automatically booting `Provider`s based on the need for a component they register, I've added a `provides` API for a `Provider` to specify what keys it intends to register once started. The `Booter` also gains `#provide?` and `#provider_for` methods, and any unresolved key will be tested against the `Booter` and any found `Provider` automatically started. 

This works for any key in general, not just root keys. The flipside is that, at least as implemented, the matching provider is only started on an exact key match, so `database.users` won't start a `Provider` with the `database` identifier in `Booter`, but if it `Inject`s `database`, then _that_ key _will_ start the `:database` provider (assuming the provider has `provides :database` in its definition.)

(edit: I'm writing acceptance tests and am seeing a lot of `register_provider(:name) { provides :name }`... and this has me thinking that `provides` maybe should default to the provider identifier, unless it is overridden. The logic would still function on the requirement that a provider knows before boot what keys it intends to register, but we'd shorthand that for the 80% use case where there is a 1:1 between provider name and registered key.

I think this also eliminates the utility of `use`, since you could just directly use the keys you expect to be there, and if there's a provider for them they'll be started implicitly. Is there a use case where you would want to ensure a provider has been started _without_ caring about its keys or looking one up? I'm not sure one way or the other.
)

### `load_component` implicit/lazy loading

I've just finished adding support to plugins for a `missing_key` hook, whereby plugins are sequentially given a chance to resolve a key that has been requested but does not yet exist on the component. This replaces the hardcoded logic to check the booter, load paths, importer, and so on. So, `Booter::Plugin` checks `Booter` to see if there's a provider that wants to provide that key, `AutoRegistrar::Plugin` checks to see if the file can be auto-registered, `Importer::Plugin` checks if the `Identified`'s root key matches an imported container, and so on, until someone returns `true` or we run out of plugins. The fallback for trying the whole process a second time but with the `Identified`'s namespace prepended to the identifier is still there in the container mixin.

### Systems

If an object mixes in `Dry::System::Mixin`, then it gains a `register_provider` class method, and will be auto-registered into the `Dry::System.systems` hash. 

## Plugins

A plugin subclasses `Dry::System::Plugin`, which doesn't do much other than add `identifier` and `reader` settings to the class. A `FooPlugin` is installed into a container via `Container.use(FooPlugin)`.

It is initialized with the container it is plugged into. In `initialize`, it can add settings to the container, `extend` it with a mixin, and so on. 

If `FooPlugin.reader` (default: `true`) is not `false`, then the value (or the `FooPlugin.identifier`, if `reader` is `true` or `nil`) is automatically transformed into a reader method name. The reader method defined on the `Container` proxies to `FooPlugin#instance`. 

If a `FooPlugin.used` class method is defined, then that'll be called as part of registering the plugin, before initialization.

Plugins are initialized at the top of `#configure`, and the `PluginManager` maintains the reference for the entire life of the app. They will respond to triggers/events if defined as instance methods, such as `#after_configure`. 

### Example: `Booter`

`Booter` is a reasonably complex example. First, it saves a reference to its container. Then it mixes in the `Container#boot` method. Mixins ideally should be kept simple with isolated methods that proxy to collaborators, instead of incorporating complex logic directly.

Finally, it pokes `Dry::System` to process any auto-registrations.

```ruby
config.identifier = :booter

def initialize(container, config)
  @container = container

  container.send(:extend, Mixin)

  Dry::System.finalize!
end
```

It instantiates an actual `Booter` on demand, initializing it with the container and the registered systems of providers. This method will be hit whenever `Container.booter` is called.

```ruby
def instance
  @instance ||= Booter.new(container, Dry::System.systems)
end
```

It listens for `after_configure`, which is special-cased to also pass along the final configuration. Then it requires all the boot files. These are not yet started, they're just registered on the `Booter` in an inert state. They won't get started automatically until finalization.

```ruby
def after_configure(config)
  boot_path = config.root.join("#{config.system_dir}/boot")

  Dir["#{boot_path.to_s}/**/#{RB_GLOB}"].map do |path|
    require(path)
  end
end
```

Finally, it listens for `finalize` and `shutdown` events.

```ruby
def finalize
  instance.finalize!
end

def shutdown
  instance.shutdown
end
```

## Minor notes

I moved the pre-existing plugins (updated to the new API, with specs) to their own directory alongside `system`, alongside `system_components`. It feels like they should be segregated from the core library somehow, but I'm not really sure the best way to do that, so I just bumped them up a level for now. I don't feel all that strongly about this, but there you have it.

In updating the specs I've removed some where I couldn't immediately tell the utility, either because they were integration tests that appeared to duplicate unit specs, or they seemed to test certain code paths redundantly, or they tested imported code. I'm thinking it might be a good idea to add a third specs directory, whether `acceptance` or `features` or whatever, for specs to go that verify documented invariants without necessarily testing specific bits of code, so that if it's important to have specs for, say, Dry::AutoInject behavior as included in Dry::System that is otherwise off-the-shelf from the gem, then those can be kept separate from tests of classes in the repo. A separate suite gets the best of both worlds: the confidence garnered from all-encompassing, general acceptance specs, and also the usability of specs that can be traced directly to implementations.


endash:
I'm going to close this, because this so-called "refactor" has taken on a life of its own, at this point. I'm still optimistic that there are things I've been working on that could merge into the original repo, but at the same time I'm in the midst of validating a significant direction change and I want to do that without the continuity constraints. 

Sometime this week I'll write up what I've been working on and post to the discussion forum, and if there's any interest we can figure out how what I've been working on can be reconciled with dry-system. 

In the mean-time, I'm leaving this branch up and the "original" description of what this PR would have entailed (even though my thinking has changed quite a bit in the last few days) for reference, or if anyone might find anything in it useful... or just for future crow-eating.

PR #108, opened 2019-03-07
Run specs in TruffleRuby on Travis CI

deepj:
All specs passed. For now, I would allow to fail to run them in TruffleRuby


solnic:
wow that's great news, ty 😄

Issue #80, opened 2018-02-10
Generate dependency graph for all objects in application container


davydovanton:
I created a simple proof of concept for generation graph of all dependencies in container: [link to gist file](https://gist.github.com/davydovanton/bd8ee3c6d0ad3e6b6f60e9ab7e539975).

For this I used ["custom" auto_inject register](https://gist.github.com/davydovanton/bd8ee3c6d0ad3e6b6f60e9ab7e539975#file-dry_system_deps-rb-L12-L20) for creating `included_keys` class method with all dependencies.

After that I [detect all classes with this method](https://gist.github.com/davydovanton/bd8ee3c6d0ad3e6b6f60e9ab7e539975#file-dry_system_deps-rb-L56-L58) and generate "AST" with associations between classes.

With this changes we can detect unused dependencies in container, build GUI graph for easily understand system (like UML diagrams) and ect. 

We can start build something like this for dry-system, WDYT? I can create a simple plan with small steps for this feature if you want.

davydovanton:
After discussion with @solnic we decided to create a graph of all registered dependencies in dry-system. Also, we want to provide public API for working with this graph. It’ll allow using `dry-system-graph` for everyone.

Our strategic product goals now: build GUI graph for dependencies, detect unused keys and calculate heat map.

## Implementation

We want to use `dry-events` and `dry-monitoring` for sending events that something was registered in the container. After that, create a new library for detecting it and detecting what and when was called. And build a graph depend on these events.

After that, we want to create plugins for GUI, unused keys and etc.

Schematically it can look like this. In dry-event we subscribe on something that happen in dry-system (registered, joined and called):

![2019-05-20 15 56 44](https://user-images.githubusercontent.com/1147484/58023259-0bdd6100-7b18-11e9-8e54-cfd95b13a456.jpg)

After that we can build a directed graph with weight for edges:

![2019-05-20 15 56 39](https://user-images.githubusercontent.com/1147484/58023273-13046f00-7b18-11e9-9819-e9c461ef3072.jpg)

## Current state and TODO
### dry-events and monitoring

- [ ] Understand how to work with `dry-events` and update (or make documentation);
- [ ] Have a call with @mensfeld do understand what our next steps in dry-monitoring library;

### dry-system dependency graph

- [ ] Update  `#register` method for broadcasting events;
- [ ] Update injectable object for `dry-system` for broadcasting events on dependency injection;
	- main issue - how to understand what was injected and target for this;
[...]

PR #113, opened 2019-05-21
Introduce dependencies events

davydovanton:
Implementation of https://github.com/dry-rb/dry-system/issues/80

## Example of usage
Screenshot 2019-05-21 at 17 20 01


lib/dry/system/plugins/dependency_graph.rb
```diff
@@ -0,0 +1,39 @@
...
+          end
+        end
+
+        # @api private
+        def self.dependencies
+          'dry/events/publisher'
+        end
+
+        def register(key, contents = nil, options = {}, &block)
+          unless key.to_sym == :notifications
```
solnic: This should be configurable as I'm pretty sure we'd like to skip it for other deps.

davydovanton: good point. will add it 👍

davydovanton:
> @davydovanton I should clarify that our rubocop rules are a WIP, so if you see something silly, please add a comment about it

got it. I think I can add local rubocop for automatic fixes in other PR and fix it. WDYT?

PR #115, opened 2019-06-04
Send register class name to instrumentation register event in DG plugin

davydovanton:
Screenshot 2019-06-04 at 17 36 55

### DIFF
--- a/lib/dry/system/plugins/dependency_graph.rb
+++ b/lib/dry/system/plugins/dependency_graph.rb
@@ -14,7 +14,7 @@ module Dry
 
           system.use(:notifications)
 
-          system.setting :ignored_dependencies, [:notifications]
+          system.setting :ignored_dependencies, []
 
           system.after(:configure) do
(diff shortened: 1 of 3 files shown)


solnic:
well that was easy 😄

lib/dry/system/plugins/dependency_graph.rb
```diff
@@ -31,12 +31,15 @@ def self.dependencies
 
         # @api private
         def register(key, contents = nil, options = {}, &block)
+          super
+
           unless config.ignored_dependencies.include?(key.to_sym)
-            notifications = self[:notifications]
-            notifications.instrument(:registered_dependency, key: key)
+            self[:notifications].instrument(
+              :registered_dependency, key: key, class_name: self[key].class
```
flash-gordon: it's not class name, it's a class

davydovanton: agree. will fix it, thanks!

PR #118, opened 2019-06-12
Make dir listings order stable

flash-gordon:
The order of result of `Dit.glob/Dir.[]` varies depending on the FS. In general, it shouldn't matter, however, in practice it may lead to random-looking crashes across different environments (like macos vs linux etc).


solnic:
this makes sense

Issue #109, opened 2019-03-12
Attempt to lazy-load components when non-finalized container is asked for `key?`


timriley:
We should copy the behaviour of `#resolve` on a non-finalized container:

```
        # @api public
        def resolve(key)
          load_component(key) unless finalized?

          super
        end
```

And do the same for `key?`. This would help unit testing of `Dry::Transaction` objects (i.e. alongside a non-finalized container), which [check with the container for a `key?` before attempting to resolve an object from it](https://github.com/dry-rb/dry-transaction/blob/08617f7f564f4871670600665b225bdf8cd1be9d/lib/dry/transaction/operation_resolver.rb#L9-L10).

flash-gordon:
resolved by #120

PR #120, opened 2019-07-08
Attempt to lazy-load component when container is asked for `key?`

flash-gordon:
This also adds `Container.registered?` which acts as `key?` used to.
Besides it, `Container.resolve` not takes an optional block to make this interface consistent with dry-container.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -2,12 +2,13 @@
 
 source 'https://rubygems.org'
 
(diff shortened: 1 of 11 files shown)


flash-gordon:
I forgot to mention, it resolves https://github.com/dry-rb/dry-system/issues/109

PR #122, opened 2019-08-30
Implement "allow" feature for injectable object

davydovanton:
Reference to https://github.com/dry-rb/dry-auto_inject/pull/59 and

### DIFF
--- a/lib/dry/system/auto_inject.rb
+++ b/lib/dry/system/auto_inject.rb
@@ -1,0 +1,44 @@
+# frozen_string_literal: true
+
+require 'dry-auto_inject'
+
+module Dry
+  module System
+    # @api private
+    module AutoInject
+      # @api private
+      ForbiddenInjectionError = Class.new(StandardError)
+
+      # Dry system specific injector builder class
+      #
+      # @api private
+      class Builder < Dry::AutoInject::Builder
+        def initialize(container, options = {})
+          super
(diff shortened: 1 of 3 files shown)


timriley:
Hi @davydovanton, thanks for your patience with this PR :) And also for showing me how it might be used when we met last month!

After thinking it over, however, I'm not sure if now's the right time to be putting such a specialised feature into dry-system itself, given that at this point we should be looking to round out the existing set of features before getting a 1.0 done.

I wonder if this particular feature can be further explored as some kind of add-on extension in the meantime? That could be good in that it'd allow further experimentation with the various ways this kind of feature could work – like I've mentioned to you before, I can certainly see how this feature is useful in certain circumstances, but I feel we could probably do a little more exploration before settling on an ideal approach. Part of me thinks that we could probably do something _richer_ than relying on simple regexp matches on string identifiers...

@davydovanton Would you be OK with taking this approach in the meantime (allowing this code to live on outside dry-system itself)?

@solnic @flash-gordon did you have any other thoughts?

timriley:
I'm closing this one for now since we haven't had active interest in it, and I think it remains quite reasonable to implement outside of dry-system (e.g. you can create an auto_inject injector with this feature, and pass the dry-system container to it directly).

After we get past 1.0 and we see (hopefully) wider usage of dry-system, we could consider this again if we see sufficient user interest :)

PR #123, opened 2019-09-22
Make dependency plugin immutable for different dry-system containers

davydovanton:
I still working on dry-system dependency graph plugin and found that current implementation use one notification. For example, we have 2 dry-system containers:

```ruby
  use :dependency_graph
  use :monitoring

  configure do |config|
    config.ignored_dependencies = %i[not_registered]
    config.auto_register = %w[lib]
    config.name = :main
  end

  load_paths!('lib')
end

class OtherApp < Dry::System::Container
  use :dependency_graph
  use :monitoring

  configure do |config|
    config.ignored_dependencies = %i[not_registered]
    config.name = :new_one
    config.auto_register = %w[]
  end
end
```

In this case, these two containers will use one `notifications` object (which was created in the last container). For prevent this behavior we need to make this plugin immutable for each container.

Also, I drop the unused code.

### DIFF
--- a/lib/dry/system/plugins/dependency_graph.rb
+++ b/lib/dry/system/plugins/dependency_graph.rb
@@ -20,7 +20,7 @@ module Dry
             self[:notifications].register_event(:resolved_dependency)
             self[:notifications].register_event(:registered_dependency)
 
-            system.strategies(Stratagies.with_notifications(system[:notifications]))
+            strategies(Strategies)
           end
         end
 
--- a/lib/dry/system/plugins/dependency_graph/strategies.rb
+++ b/lib/dry/system/plugins/dependency_graph/strategies.rb
@@ -5,19 +5,8 @@ module Dry
     module Plugins
       module DependencyGraph
         # @api private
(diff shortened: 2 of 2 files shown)


lib/dry/system/plugins/dependency_graph.rb
```diff
@@ -20,7 +20,7 @@ def self.extended(system)
             self[:notifications].register_event(:resolved_dependency)
             self[:notifications].register_event(:registered_dependency)
 
-            system.strategies(Stratagies.with_notifications(system[:notifications]))
+            strategies(Stratagies)
```
kaikuchn: Is that a typo? If so, it's also in the `lib/dry/system/plugins/dependency_graph/strategies.rb`

```suggestion
            strategies(Strategies)
```

davydovanton: hey, yep. My bad. thanks for finding this

solnic:
@davydovanton this makes sense

PR #125, opened 2019-10-29
Accept proc for root setting

shanecav84:
This allows a gem to use dry-system and set the root relative to the gem and not the software using the gem.

Per [previous discussion](https://discourse.dry-rb.org/t/3rd-party-bootable-component/505), I have the same requirement. I would also like to use `dry-system` to boot my gem. This PR allows me to set the proper root for a gem. Otherwise, the root can only be set relative to the software using the gem. For example:

```ruby
# /my/gem
module MyGem
    configure do |config|
    config.root = -> { 'root' }
  end
end

# /somewhere/else
require 'my_gem'

MyContainer.root
# Before
#=> /somewhere/else/root
# After
#=> /my/gem/root
```

### DIFF
--- a/lib/dry/system/container.rb
+++ b/lib/dry/system/container.rb
@@ -74,7 +74,10 @@ module Dry
 
       setting :name
       setting :default_namespace
-      setting(:root, Pathname.pwd.freeze) { |path| Pathname(path) }
+      setting(:root, Pathname.pwd.freeze) do |path|
+        path = path.is_a?(Proc) ? path.call : path
+        Pathname(path)
+      end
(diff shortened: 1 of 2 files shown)


timriley:
Hi @shanecav84, so does this not work?

```ruby
module MyGem
    configure do |config|
    config.root = __dir__
  end
end
```

In my understanding of `__dir__` (and also `__FILE__`), they return a path based on the _file in which they were called_, which in your case would be the file within your gem's own directory structure.

shanecav84:
That's working! Thanks.

PR #134, opened 2020-01-13
Fail fast if auto_registrar config contains incorrect path

cutalion:
Check that the directory in the `config.auto_register = %w[dir]` or `container.auto_register!('dir')` exists and fail if it's not.

I found it a bit confusing that `auto_register!('persistence')` didn't warn me that it should be `lib/persistence`.


solnic:
Thank you! This is definitely a nice improvement.

PR #135, opened 2020-01-19
Misspelled plugin name raises meaningful error

### LINKED ISSUE #132: Misspelled plugin name causes confusing exception
**Describe the bug**

When a plugin name is misspelled, you'll end up with a confusing exception.

**To Reproduce**

```ruby
  use :roargh
end
```

Results in:

```
NoMethodError:
  undefined method `load_dependencies' for nil:NilClass
# .gem/ruby/2.4.1/gems/dry-system-0.12.0/lib/dry/system/plugins.rb:86:in `use'
```

**Expected behavior**

A meaningful error should be raised saying `"Plugin :roargh does not exist"`

cgeorgii:
This pull request fixes #132.

**Note:** I wasn't sure where to put the new `Dry::System::PluginNotFoundError` definition. There are two files that contain error definitions, `lib/dry/system/constants.rb` and `lib/dry/system/errors.rb` and the latter seemed like the correct place. In fact, I'm not really sure whether it makes sense to have error definitions in the constants file, so let me know if I can move them.

### DIFF
--- a/lib/dry/system/errors.rb
+++ b/lib/dry/system/errors.rb
@@ -75,5 +75,14 @@ module Dry
         super("component +#{component_name}+ has not been started")
       end
     end
+
+    # Error raised when trying to use a plugin that does not exist.
+    #
+    # @api public
+    PluginNotFoundError = Class.new(StandardError) do
+      def initialize(plugin_name)
+        super("Plugin #{plugin_name.inspect} does not exist")
+      end
+    end
   end
 end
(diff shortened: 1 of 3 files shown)


solnic:
Thanks!

> Note: I wasn't sure where to put the new Dry::System::PluginNotFoundError definition. There are two files that contain error definitions, lib/dry/system/constants.rb and lib/dry/system/errors.rb and the latter seemed like the correct place. In fact, I'm not really sure whether it makes sense to have error definitions in the constants file, so let me know if I can move them.

Yes, exception classes should go to `errors.rb`. We should move stuff from `constants.rb`.

PR #139, opened 2020-02-03
Add args and hash strategies from auto inject to dependency graph plugin

davydovanton:
Hey, I forgot to add other strategies to dry-system dependency graph. We need to do it if we want to detect injectors for graph

### DIFF
--- a/lib/dry/system/plugins/dependency_graph/strategies.rb
+++ b/lib/dry/system/plugins/dependency_graph/strategies.rb
@@ -21,7 +21,34 @@ module Dry
             end
           end
 
+          # @api private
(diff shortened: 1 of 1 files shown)


timriley:
@davydovanton This change looks 👍 to me.

I wonder why the tests are failing, though... @flash-gordon would you have any idea?

solnic:
rebase and it’ll pass on CI

davydovanton:
@solnic done, thanks for help!

solnic:
@davydovanton could you expand `spec/integration/container/plugins/dependency_graph_spec.rb` to cover this addition?

PR #143, opened 2020-02-25
Fix error with mutation of frozen container

davydovanton:
Hey,

@vessi found an interesting bug related to `strategies` logic. You can reproduce this bug with this script:

```ruby

end

class Container2 < Dry::System::Container
  strategies # if you call empty `strategies` everything will be okay because we will mutate state from a not frozen container
end

Container.finalize!
Container2.finalize!

# We already mutate state in class method `strategies``
Import2 = Container2.injector
# Will fail
Import = Container.injector
```

If you call `injector` before `finalize!` everything will be okay. But if not you get an error message:
```
Traceback (most recent call last):
        2: from test.rb:15:in `<main>'
        1: from /Users/anton/.rvm/gems/ruby-2.5.0/bundler/gems/dry-system-6ff22632ebc2/lib/dry/system/container.rb:491:in `injector'
/Users/anton/.rvm/gems/ruby-2.5.0/bundler/gems/dry-system-6ff22632ebc2/lib/dry/system/container.rb:94:in `strategies': can't modify frozen #<Class:Container> (FrozenError)
```

As you can see this bug related to frozen container and `@strategies` mutation from this line: https://github.com/dry-rb/dry-system/blob/master/lib/dry/system/container.rb#L94

I created a patch which works well:

**Before patch**

Screenshot 2020-02-26 at 01 16 13

**After patch**

Screenshot 2020-02-26 at 01 33 19

------

I have only one problem with this code. Now dry-system spam with `warning: instance variable @strategies not initialized` warning and I have no idea how I can fix it in the current implementation. Also, I'll be happy if you help me with it

### DIFF
--- a/lib/dry/system/container.rb
+++ b/lib/dry/system/container.rb
@@ -90,8 +90,10 @@ module Dry
         def strategies(value = nil)
           if value
             @strategies = value
+          elsif defined?(@strategies) && @strategies
+            @strategies
           else
-            @strategies ||= Dry::AutoInject::Strategies
+            Dry::AutoInject::Strategies
           end
         end
 
--- a/spec/unit/container_spec.rb
+++ b/spec/unit/container_spec.rb
@@ -308,4 +308,18 @@ RSpec.describe Dry::System::Container do
       expect(container.registered?('test.dep')).to be(true)
     end
   end
+
+  describe '.injector' do
+    subject(:container) { Test::InjectorContainer }
+
+    before do
+      class Test::InjectorContainer < Dry::System::Container
+      end
+      Test::InjectorContainer.finalize!
+    end
+
+    it 'checks if a component is registered' do
+      expect { container.injector }.not_to raise_error
+    end
+  end
 end


lib/dry/system/container.rb
```diff
@@ -91,7 +91,7 @@ def strategies(value = nil)
           if value
             @strategies = value
           else
-            @strategies ||= Dry::AutoInject::Strategies
+            @strategies || Dry::AutoInject::Strategies
```
timriley:
```suggestion
            if instance_variable_defined?(:@strategies) && @strategies
              @strategies
            else
              Dry::AutoInject::Strategies
            end
```

This should fix your undefined ivar warnings, @davydovanton?

timriley: What I would probably do, though, is flatten everything into an if/elsif/else

```ruby
if value
  @strategies = value
elsif instance_variable_defined?(:@strategies) && @strategies
  @strategies
else
  Dry::AutoInject::Strategies
end
```

davydovanton: thanks for the suggestion! I selected the second approach (for skipping nesting condition) 💯

solnic:
> Longer term there might be other ways to address this, e.g. adding an actual container-level setting for "injector options", which would then get handled/frozen properly along with all the other settings, but something like that should be addressed in a different PR.

This *should be a setting*, so `setting :injector_strategies, AutoInject::Strategies, reader: true` - I'm not sure how we ended with a get-or-set method?

davydovanton:
@solnic It was a hack for faster prototyping. I think that the idea of the setting is great. I can update this PR or we can merge fix and refactor this option. I prefer the second option but WDYT about it?

solnic:
@davydovanton it'd be good to change it to be a setting in this PR already because right now it's incosistent.

timriley:
This is taken care of now by #210. I've taken a slightly different approach than the suggestion above of having the injector strategies be a setting, but it still leaves that option open for us if in the future we decide it makes more sense.

PR #144, opened 2020-03-17
Document require before plugin use

adam12:
The `dry/system/components` file must be required for `dry-system` plugins to be usable.


adam12:
Not required.

PR #145, opened 2020-04-06
Add information about built-in settings system

davydovanton:
I understood that we don't have a documentation page about settings in dry-system. That's why I created this PR. I know that it's a raw page and we can make it better in the future. But I think it will be a good start if we merge it asap and update it in the next iteration. WDYT?

@solnic also, a question about documentation: should I update something else or it's enough?

### DIFF
--- a/docsite/source/settings.html.md
+++ b/docsite/source/settings.html.md
@@ -44,7 +44,7 @@ Application.boot(:redis) do |container|
 end
 ```
 
-Or use settings in your logic:
+Or use `:settings` as an injectible dependency in your classes:
 
 ```ruby
   module Operations
--- a/docsite/source/settings.html.md
+++ b/docsite/source/settings.html.md
@@ -6,7 +6,7 @@ name: dry-system
 
 ## Basic usage
 
-Dry-system provides a built-in `:settings` component which you can use in your application. In order to set it up, simply define a bootable `:settings` component and import it from the `:system` provider:
+dry-system provides a built-in `:settings` component which you can use in your application. In order to set it up, simply define a bootable `:settings` component and import it from the `:system` provider:
 
 ```ruby
 # in system/boot/settings.rb
--- a/docsite/source/settings.html.md
+++ b/docsite/source/settings.html.md
@@ -71,4 +71,4 @@ settings do
 end
 ```
 
-In this case, if you don't have `ENV['REDIS_URL']` value you get `''` as a default value for `settings.redis_url` calls
+In this case, if you don't have `ENV['REDIS_URL']` value, you get `''` as the default value for `settings.redis_url` calls.
--- a/docsite/source/settings.html.md
+++ b/docsite/source/settings.html.md
@@ -6,7 +6,7 @@ name: dry-system
 
 ## Basic usage
 
-Dry-system has already built-in settings system which you can use for your application. For this you need to use system component:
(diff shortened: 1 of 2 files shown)


docsite/source/settings.html.md
```diff
@@ -0,0 +1,74 @@
+---
+title: Settings
+layout: gem-single
+name: dry-system
+---
+
+## Basic usage
+
+Dry-system has already built-in settings system which you can use for your application. For this you need to use system component:
```
solnic, line 9:
```suggestion
`dry-system` provides a built-in `:settings` component which you can use in your application. In order to set it up, simply define a bootable `:settings` component and import it from the `:system` provider:
```

docsite/source/settings.html.md
```diff
@@ -0,0 +1,74 @@
...
+
+    uri = URI.parse(container[:settings].redis_url)
+    redis = Redis.new(host: uri.host, port: uri.port, password: uri.password)
+
+    container.register('persistance.redis', redis)
+  end
+end
+```
+
+Or use settings in your logic:
```
solnic, line 47:
```suggestion
Or use `:settings` as an injectible dependency in your classes:
```

docsite/source/settings.html.md
```diff
@@ -0,0 +1,74 @@
...
+
+You can [use dry-type](https://dry-rb.org/gems/dry-types/master/default-values/) for provide default value for specific setting:
+
+```ruby
+settings do
+  key :redis_url, Types::Coercible::String.default('')
+end
+```
+
+In this case, if you don't have `ENV['REDIS_URL']` value you get `''` as a default value for `settings.redis_url` calls
```
solnic, line 74:
```suggestion
In this case, if you don't have `ENV['REDIS_URL']` value, you get `''` as the default value for `settings.redis_url` calls.
```

solnic:
> should I update something else or it's enough?

@davydovanton this is enough, once this is merged a workflow will ping dry-rb.org repo to build&deploy

docsite/source/settings.html.md
timriley: @solnic do you mind if we don’t use `code-style` for our gem names in docs? They’re not technically code, they’re (unconventionally lower-cased) proper nouns.

solnic: @timriley sure, but I cringe when I see `Dry-foo` and OTOH starting a sentence with "dry-system" looks wrong, so I dunno... 😅

timriley: I think it’s ok to start a sentence with the lower-case, normally styled text :)

solnic: @timriley ok great then, I'm totally ok with dropping "`" then :)

davydovanton:
@solnic @timriley thanks for the review. Fixed everything 💯

docsite/source/settings.html.md
```diff
@@ -0,0 +1,74 @@
+---
+title: Settings
+layout: gem-single
+name: dry-system
+---
+
+## Basic usage
+
+Dry-system provides a built-in `:settings` component which you can use in your application. In order to set it up, simply define a bootable `:settings` component and import it from the `:system` provider:
```
solnic, line 9:
```suggestion
dry-system provides a built-in `:settings` component which you can use in your application. In order to set it up, simply define a bootable `:settings` component and import it from the `:system` provider:
```

Tim said it's OK to use the actual, lower-cased, gem name even in the beginning of a sentence 😄

davydovanton: 😄

PR #151, opened 2020-08-15
Support multiple boot file dirs

timriley:
This PR adds a `bootable_dirs` setting to `Dry::System::Container` and updates the `Booter` so that it finds boot files from one or more directories.

For each configured `bootable_dirs` entry, relative directories will be appended to the configured container root, and absolute directories will be left unchanged.

When searching for bootable files, the first match will win, and any subsequent same-named files will not be loaded. In this way, the bootable_dirs acts similarly to the $PATH in a shell environment.

This arrangement will support uses where a container vendor may want to ship their own “default” bootable components, but allow the _user_ of the container to replace any of them one by one. To achieve this, the user can add their own entry to the beginning of the bootable_dirs setting, which will allow their own version of the bootable components to take precedence.

---

**Notes for reviewers:**

- If you check the commit-by-commit history, you'll note I did a little tidying up before committing the actual new functionality here
- As part of adjusting the `Booter`, I also moved some truly internal methods to be under a `private` section of the class.
- I still find parts of our bootable component functionality particularly hairy, but I'll save a refactoring of those for a future PR :)

### DIFF
--- a/lib/dry/system/booter.rb
+++ b/lib/dry/system/booter.rb
@@ -129,11 +129,6 @@ module Dry
       end
 
       # @api private
-      def lifecycle_container(container)
-        LifecycleContainer.new(container)
-      end
-
-      # @api private
       def with_component(id_or_component)
         component =
           case id_or_component
--- a/lib/dry/system/booter.rb
+++ b/lib/dry/system/booter.rb
@@ -120,7 +120,7 @@ module Dry
       # @api private
       def call(name_or_component)
         with_component(name_or_component) do |component|
-          raise ComponentFileMismatchError.new(name, registered_booted_keys) unless component
+          raise ComponentFileMismatchError.new(name) unless component
 
           yield(component) if block_given?
 
--- a/lib/dry/system/booter.rb
+++ b/lib/dry/system/booter.rb
@@ -5,6 +5,7 @@ require "dry/system/errors"
 require "dry/system/constants"
 require "dry/system/lifecycle"
 require "dry/system/booter/component_registry"
+require "pathname"
 
 module Dry
   module System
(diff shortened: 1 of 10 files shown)


lib/dry/system/booter.rb
```diff
@@ -16,29 +16,38 @@ module System
...
+        # path.join("#{name}#{RB_EXT}")
+
+        find_boot_file(name)
+      end
+
+      def find_boot_file(name)
+        paths.detect { |path|
+          file_name = path.join("#{name}#{RB_EXT}")
+          break file_name if File.exist?(file_name)
+        }
```
solnic, line 50: @timriley this should raise an exception if it finds > 1 file

lib/dry/system/container.rb
```diff
@@ -75,7 +73,7 @@ class Container
       setting :name
       setting :default_namespace
       setting(:root, Pathname.pwd.freeze) { |path| Pathname(path) }
-      setting :system_dir, "system"
+      setting :bootable_dirs, ["system/boot"]
```
solnic, line 76: @timriley so, this seems to be a rather serious API breakage. The system dir was meant to be used as a place for keeping core system files, like container definition and auto-inject module. I'm not sure why removing this feature is needed for the bootable refactor. Could this at least be separated in another PR? I'm not against removing it btw, I just need to understand why we're a) removing it here b) hardcoding `system/boot` whereas previously it could be configured

timriley: @solnic yeah, you’re right, I should pare this PR back to supporting multiple boot dirs only. 

I do think that the system dir setting _could_ be removed before 1.0 but we can definitely have a separate conversation about that.

lib/dry/system/booter.rb
timriley: @solnic this is actually expressly designed to support different boot dirs having files with the same name. The idea is that this is like a shell $PATH, so you can prepend your own bootable first to provide a customized replacement of a bootable component that you’d find within a later entry in the dirs. 

See the attached Hanami PR for a demonstration of why I want it to be that way — Hanami can provide its own standard boot files, but still allow app authors to swap them out for their own implementations on a fine-grained basis.

lib/dry/system/container.rb
solnic: @timriley sounds good 🙂

v-kolesnikov:
Just noticed that this change is breaking in case if you previously had a component in subdirectory ('system/boot/perstince/database.rb for example) as:

```ruby
MyApp::Container.boot :database, namespace: :persistence do
  init do
    register :database, :database
  end
end
```

and the container:

```ruby
require 'bundler/inline'

gemfile(true) do
  gem 'dry-system', '0.18.1'
end

module MyApp
    configure do
      config.name = :my_app
    end
  end
end

MyApp::Container.finalize!

p MyApp::Container['persistence.database']
```

It won't work on 0.18.1. Adding
```ruby
      config.bootable_dirs = [
        'system/boot/**'
      ]
```
to container fixes the issue.

@timriley JFYI

PR #153, opened 2020-11-11
Add autoloader support, clarify load path handling

timriley:
This PR adds support for working with autoloaders like [Zeitwerk](http://github.com/fxn/zeitwerk).

It achieves this by:

- Moving the responsibility for requiring files away from `Container` (removing `.require_path`, **a breaking change**) and into the `Component` (as `Component#require!`), which itself delegates to the configured `Loader` (via `Loader#require!)
- Adding a new `Dry::System::Loader::Autoloading` loader, which can be configured to replace the default loader. This is identical to the default loader except that instead of requiring the component's file, it accesses the component's constant, which is a sufficient trigger for an autoloader like Zeitwerk to load the constant based on its own configuration

Along with this, it makes some internal changes and to rationalise some related underpinnings:

- Renames `Container.load_paths!` to `.add_to_load_path!`, to make its function much clearer (**this is a breaking change**)
- Stops automatically adding the `system_dir` to the load path, since its now a reasonable use case to run dry-system without any of its managed directories being on the load path (**this is a breaking behavioural change**)
- Adds a new `component_dirs` setting, defaulting to `["lib"]`, which is used to verify whether a given component is "local" to the container. This check was previously done using the directories previously passed to `load_paths!`, which we can't rely upon now that we're supporting autoloaders.
- Adds a new `add_component_dirs_to_load_path` setting, defaulting to `true`, which will automatically add the configured `component_dirs` to the load path in an after-configure hook. This will help ease the transition from the previous behaviour, and make dry-system still work nicely when _not_ using an autoloader.

To see this all in action, check out the integration test using Zeitwerk: `spec/integration/container/autoloading_spec.rb`. 

Here's the highlights, for easy reference. Container setup:

```ruby
module Test
    config.root = SPEC_ROOT.join("fixtures/autoloading").realpath
    config.add_component_dirs_to_load_path = false
    config.loader = Dry::System::Loader::Autoloading
    config.default_namespace = "test"
  end
end
```

Zeitwerk setup:

```ruby
loader = Zeitwerk::Loader.new
loader.push_dir Test::Container.config.root.join("lib").realpath
loader.setup
```

Then given a component "foo", at `lib/test/foo.rb`:

```ruby
module Test
  class Foo
    def call
      # Referencing this constant without a require!
      Entities::FooEntity.new
    end
  end
end
```

Resolving `Test::Container["foo"]` will give you an instance of `Test::Foo` as expected, and then a `#call` will also return an `Entities::FooEntity` as expected. Too easy!

---

This PR is the first step of several. The next step will be to add a "rich configuration" API for adding component dirs, and allowing auto-registration to be enabled/disabled/configured independently for each component dir. With this in place, we can then remove the `auto_register` setting (which no longer makes sense as an independent setting, since the auto-registrar should _only_ be operating on component dirs), as well as the `auto_register!` method (for the same reasons).

A further step could be to add a `zeitwerk` _plugin_ to dry-system (just like how we have a `bootsnap` plugin) that takes care of setting up the zeitwerk loader for the configured component dirs, as well as configuring the autoloading loader, which would give dry-system users a simple one-line way to enable zeitwerk!

At this point (or perhaps before doing the plugin) I'll switch back to Hanami and work on configuring its container to work with zeitwerk, which will be a useful check to ensure this new arrangement in dry-system is offering everything we need. Once that is done, I think we can work on cutting a new dry-system release. (This will also need some coordination with dry-rails to make sure we don't break it)

Time-wise, I hope to get most of this done across the month of November.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -1,14 +1,15 @@
 # frozen_string_literal: true
 
-source 'https://rubygems.org'
+source "https://rubygems.org"
 
-eval_gemfile 'Gemfile.devtools'
+eval_gemfile "Gemfile.devtools"
 
 gemspec
 
-gem 'bootsnap'
-gem 'dry-monitor'
+gem "bootsnap"
+gem "dry-monitor"
+gem "zeitwerk"
 
 group :tools do
-  gem 'pry-byebug', platforms: :mri
+  gem "pry-byebug", platforms: :mri
 end
--- a/Gemfile
+++ b/Gemfile
@@ -6,7 +6,8 @@ eval_gemfile "Gemfile.devtools"
 
 gemspec
 
-gem "bootsnap"
+# Remove verson constraint once latter versions release their -java packages
+gem "bootsnap", "= 1.4.9"
 gem "dry-monitor"
 gem "zeitwerk"
 
--- a/lib/dry/system/component.rb
+++ b/lib/dry/system/component.rb
@@ -93,6 +93,10 @@ module Dry
         freeze
       end
 
+      def require!
+        loader.require!
+      end
+
       # Returns components instance
       #
       # @example
--- a/lib/dry/system/container.rb
+++ b/lib/dry/system/container.rb
@@ -77,6 +77,7 @@ module Dry
       setting :bootable_dirs, ["system/boot"]
       setting :registrations_dir, "container"
       setting :component_dirs, ["lib"]
+      setting :add_component_dirs_to_load_path, true
       setting :auto_register, []
       setting :inflector, Dry::Inflector.new
       setting :loader, Dry::System::Loader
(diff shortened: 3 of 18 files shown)


timriley:
So this failed on 2.5.8 in CI. I've run rspec bisect locally and reduced the failing run to:

```
rspec ./spec/unit/container/auto_register_spec.rb[1:3:2:1:1] ./spec/unit/loader/autoloading_spec.rb[1:1:1] --seed 9224
```

Confirmed this fails on 2.5.8 locally, but passes on 2.7.2.

Will have to look into why it's happening, but open to any wisdom you might have to share in the meantime :)

solnic (approved):
LGTM!

jodosha (approved):
@timriley This looks great! 👏

Bertg:
@timriley Is there any timeline as to when we can expect this to be released?

timriley:
@Bertg I expect sometime in the next couple of months. There are breaking changes here and I need to finish off a range of related things (see #155 for starters) that make sense to release alongside it.

PR #154, opened 2020-11-26
Add dry-singleton

hguerrerojaime:


solnic:
Thanks for the PR but I can't accept it without understanding what problem it tries to solve. Please open a thread on discourse.dry-rb.org and describe your problem there first. I have a feeling we don't need to add anything to dry-system to solve your problem.

hguerrerojaime:
The issue that I faced was when I enabled stubs in the dry container, when I used the auto-injector property, the mocks were very inconsistent because the injector just set the attributes once per class when I use a singleton class, I've seen a lot of issues requesting to "memoize" auto scanned classes in the dry system, and this will solve that.

solnic:
@hguerrerojaime this should be fixed in the stub feature rather than through a custom singleton implementation, which was a workaround. I'm aware of various issues with the stub feature btw, this will be *greatly* improved before 1.0.0. Feel free to report an issue describing the problems you've faced when using `Container.stub`.

PR #155, opened 2020-12-01
Add rich component dirs config

timriley:
This adds a "rich" `component_dirs` setting, which allows multiple component directories to be added and configured independently. For example:

```ruby
  configure do |config|
    config.root = __dir__

    config.component_dirs.add "lib" do |dir|
      dir.auto_register = true    # defaults to true
      dir.add_to_load_path = true # defaults to true
      dir.default_namespace = "my_app"
    end
    
    config.component_dirs.add "another" do |dir|
      # etc.
    end
  end
end
```

The following top-level settings have been removed:

- `default_namespace`
- `add_component_dirs_to_load_path`
- `auto_register`

Since these are now configurable per-component directory, they no longer make sense at the top-level (and really, they only mostly worked by happenstance until now, relying upon the fact that most usages of dry-system involved a single primary directory for auto-registration).

The `component_dirs` setting value is a custom object, a `Dry::System::Config::ComponentDirs`, which yields a `Dry::System::Config::ComponentDir` for each added directory. These are simple classes including `Dry::Configurable` and delegating all methods to the internal `config` object via `#method_missing` (sidenote: I've followed this pattern in so many places now I think dry-configurable should support it out of the box). The `ComponentDirs` setting value is made cloneable (which is required so that we don't needlessly propagate component dir configs across different containers) via a new `cloneable` option on dry-configurable settings (see https://github.com/dry-rb/dry-configurable/pull/102, which I'll release after some final testing of this branch of dry-system).

The `component_dirs` are then exposed as `Container.component_dirs`, which are instances of `Dry::System::ComponentDir`, representing the component directory placed within the container's configured `root`. These objects wrap the component dir's config, and importantly, provide a `#component_file(component_path)` method for finding component files (i.e. Ruby source files) within the directory.

This method is then used by another new method, `Component.locate(identifier, component_dirs, **options)`, called from `Container.component`, which is the sole method for the container to build `Component` instances. `.locate` is passed the container's array of `component_dis`, and will return a `Component` instance for the first matching component file (in this way, component dirs act like a typical "path", in that the first match wins).

The benefit of this arrangement is that because `Component.locate` now has the full set of `ComponentDir` objects, which contain all their configuration, then the resulting `Component` instance can access that configuration as required! In particular, this means a component found within a `component_dir` configured with a `default_namespace` can be initialized with that namespace. This allows us to greatly simplify `Container.load_local_component` (more on that later) and reduce it from a two-pass operation (in which the method would call itself again with the `default_namespace` prepended if a component could not be found in the first pass) to a single-pass.

With this change, components are found directly from the file system in _both_ main container use cases, (a) when finalizing the container and auto-registering all components, as well as (b) when components are lazily loaded in a non-finalized container. This allows us to provide a `file_path` arg when initializing a `Component`, which means every component can now be asked if its `#file_exists?`. This new consistency across components allows us to further simplify `.load_local_component` by doing away with the entire `.require_component` block method, whose only jobs were to (a) try and find a file matching the already-initialized component (no longer required, we know about the file up front now!), and (b) require that file (also no longer required, since this PR has `Loader#call` make a call to `require!` as a first step, keeping in mind that `require!` may be a no-op for the autoloading loader). This change reduces the amount of low-level implementation detail that `Container` needs to know about, and pushes that detail down into the place where it is most relevant, the `Loader`.

(A small supporting change to `Loader` is that it's now initialized with the whole `component`, rather than just the component's path. This allows it to check if the `component.file_exists?` before attempting to require it)

Another way we increase the consistency across lazily-loaded vs auto-registered components is by adding a check for a file's magic comment-based configuration as a step inside `Component.locate`. This means that if a component has a magic comment specifying `auto_register: false`, then it can also be skipped when lazily resolving, which was never the case before! I plan to take this further and completely remove the magic comment parsing from `AutoRegistrar`, but that will happen when I remove `Container.auto_register!` in a follow-up PR (this current PR is already big enough!).

A big upshot of all these changes is that they allow `Container.load_component` and `.load_local_component` to be dramatically simplified! `.load_component` now contains _all_ the detail of deciding which loading mechanism to use. Instead of 4 levels of nesting, it now has only 1 level, with two early returns for the cases where a bootable component has already done the work of loading the specified component. `.load_local_component` drops from a 14-line method with multiple conditionals (and lots of overlap with `.load_component`) to a 3-line method with the simple job of registering the component or not. This pair of methods has confused me nearly every time I've encountered them over the last few years, and I'm so happy the changes for the `component_dirs` made this refactoring possible!

This PR has **BREAKING CHANGES**:

- The removal of the `default_namespace`, `add_component_dirs_to_load_path`, `auto_register` top-level settings from `Dry::System::Container`
- The removal of `Dry::System::ComponentLoadError`, which was previously only raised in a very particular circumstance, when attempting to lazily load a local component, but failing to find its file. Now we let a `Dry::Container::Error` bubble up like all other cases of attempting to resolve a component that doesn't exist. The custom exception didn't reveal anything useful in its error message, and it's better to be consistent here, I think.

As for next steps here:

- I'd love your review of this PR. It's an important change, so please take the time you need.
- I'll test these changes with one of the larger dry-system-backed production apps I have access to
- If that goes well, I'll release the dry-configurable changes in https://github.com/dry-rb/dry-configurable/pull/102 and update this branch to used the released version
- Once all reviewers are happy, we can merge this PR

There's one follow-up change I'd like to make before this can be released in a new gem version, and that's to rip out `Container.auto_register!` given all auto-registration should now happen only over the configured `component_dirs`. This requires some changes to dry-rails, so I'd like to do that as a separate PR.

### DIFF
--- a/lib/dry/system/auto_registrar.rb
+++ b/lib/dry/system/auto_registrar.rb
@@ -26,7 +26,9 @@ module Dry
 
       # @api private
       def finalize!
-        Array(config.auto_register).each { |dir| call(dir) }
+        config.component_dirs.each do |dir|
+          call(dir.path) if dir.auto_register
+        end
       end
 
       # @api private
@@ -36,11 +38,9 @@ module Dry
         components(dir).each do |component|
           next if !component.auto_register? || registration_config.exclude.(component)
 
-          container.require_component(component) do
-            register(component.identifier, memoize: registration_config.memoize) {
-              registration_config.instance.(component)
-            }
-          end
+          register(component.identifier, memoize: registration_config.memoize) {
+            registration_config.instance.(component)
+          }
         end
       end
 
--- a/lib/dry/system/auto_registrar.rb
+++ b/lib/dry/system/auto_registrar.rb
@@ -3,6 +3,7 @@
 require "dry/system/constants"
 require "dry/system/magic_comments_parser"
 require "dry/system/auto_registrar/configuration"
+require_relative "component"
 
 module Dry
   module System
@@ -34,11 +35,13 @@ module Dry
       # @api private
       def call(dir)
         registration_config = Configuration.new
-        yield(registration_config) if block_given?
-        components(dir).each do |component|
-          next if !component.auto_register? || registration_config.exclude.(component)
 
-          register(component.identifier, memoize: registration_config.memoize) {
+        yield(registration_config) if block_given?
+
+        components(dir).each do |component|
+          next unless register_component?(component, registration_config)
+
+          container.register(component.identifier, memoize: registration_config.memoize) {
             registration_config.instance.(component)
           }
         end
@@ -46,17 +49,18 @@ module Dry
 
       private
 
-      # @api private
       def components(dir)
-        files(dir)
-          .map { |file_name| [file_name, file_options(file_name)] }
-          .map { |file_name, options| component(relative_path(dir, file_name), **options) }
-          .reject { |component| registered?(component.identifier) }
+        files(dir).map { |file_name|
+          Component.new(
+            relative_path(dir, file_name),
+            file_path: file_name,
+            **MagicCommentsParser.(file_name)
+          )
+        }
       end
 
-      # @api private
       def files(dir)
-        components_dir = File.join(root, dir)
+        components_dir = File.join(container.root, dir)
 
         unless ::Dir.exist?(components_dir)
           raise ComponentsDirMissing, "Components dir '#{components_dir}' not found"
@@ -65,35 +69,15 @@ module Dry
         ::Dir["#{components_dir}/**/#{RB_GLOB}"].sort
       end
 
-      # @api private
       def relative_path(dir, file_path)
-        dir_root = root.join(dir.to_s.split("/")[0])
+        dir_root = container.root.join(dir.to_s.split("/")[0])
         file_path.to_s.sub("#{dir_root}/", "").sub(RB_EXT, EMPTY_STRING)
       end
 
-      # @api private
-      def file_options(file_name)
-        MagicCommentsParser.(file_name)
-      end
-
-      # @api private
-      def component(path, **options)
-        container.component(path, **options)
-      end
-
-      # @api private
-      def root
-        container.root
-      end
-
-      # @api private
-      def registered?(name)
-        container.registered?(name)
-      end
-
-      # @api private
-      def register(*args, &block)
-        container.register(*args, &block)
+      def register_component?(component, registration_config)
+        !container.registered?(component) &&
+          component.auto_register? &&
+          !registration_config.exclude.(component)
       end
     end
   end
--- a/lib/dry/system/booter.rb
+++ b/lib/dry/system/booter.rb
@@ -41,6 +41,20 @@ module Dry
         self
       end
 
+      # Returns a bootable component if it can be found or loaded, otherwise nil
+      #
+      # @return [Dry::System::Components::Bootable, nil]
+      # @api private
+      def find_component(identifier)
+        return components[identifier] if components.exists?(identifier)
+
+        return if finalized?
+
+        require_boot_file(identifier)
+
+        components[identifier] if components.exists?(identifier)
+      end
+
       # @api private
       def finalize!
         boot_files.each do |path|
@@ -53,6 +67,13 @@ module Dry
 
         freeze
       end
+
+      # @!method finalized?
+      #   Returns true if the booter has been finalized
+      #
+      #   @return [Boolean]
+      #   @api private
+      alias_method :finalized?, :frozen?
 
       # @api private
       def shutdown
@@ -168,9 +189,10 @@ module Dry
         self
       end
 
+      # TODO: this name should reflect we're doing a root_key check
+      # TODO: find out in what situations we're actually passing something with a .root_key
       def boot_file(name)
         name = name.respond_to?(:root_key) ? name.root_key.to_s : name
-
         find_boot_file(name)
       end
 
--- a/lib/dry/system/booter.rb
+++ b/lib/dry/system/booter.rb
@@ -31,11 +31,6 @@ module Dry
       end
 
       # @api private
-      def bootable?(component)
-        !boot_file(component).nil?
-      end
-
-      # @api private
       def register_component(component)
         components.register(component)
         self
@@ -136,9 +131,11 @@ module Dry
 
       # @api private
       def boot_dependency(component)
-        boot_file = boot_file(component)
+        name = component.respond_to?(:root_key) ? component.root_key : component.to_sym
 
-        start(boot_file.basename(".*").to_s.to_sym) if boot_file
+        if (component = find_component(name))
+          start(component)
+        end
       end
 
       # Returns all boot files within the configured paths
@@ -187,13 +184,6 @@ module Dry
         Kernel.require path unless components.exists?(identifier)
 
         self
-      end
-
-      # TODO: this name should reflect we're doing a root_key check
-      # TODO: find out in what situations we're actually passing something with a .root_key
-      def boot_file(name)
-        name = name.respond_to?(:root_key) ? name.root_key.to_s : name
-        find_boot_file(name)
       end
 
       def require_boot_file(identifier)
--- a/lib/dry/system/booter.rb
+++ b/lib/dry/system/booter.rb
@@ -120,7 +120,14 @@ module Dry
         start(boot_file.basename(".*").to_s.to_sym) if boot_file
       end
 
-      # @api private
+      # Returns all boot files within the configured paths
+      #
+      # Searches for files in the order of the configured paths. In the case of multiple
+      # identically-named boot files within different paths, the file found first will be
+      # returned, and other matching files will be discarded.
+      #
+      # @return [Array<Pathname>]
+      # @api public
       def boot_files
         @boot_files ||= paths.each_with_object([[], []]) { |path, (boot_files, loaded)|
           files = Dir["#{path}/#{RB_GLOB}"].sort
--- a/lib/dry/system/component.rb
+++ b/lib/dry/system/component.rb
@@ -7,6 +7,7 @@ require "dry/inflector"
 require "dry/system/loader"
 require "dry/system/errors"
 require "dry/system/constants"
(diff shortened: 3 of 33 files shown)


lib/dry/system/auto_registrar.rb
```diff
@@ -34,9 +35,15 @@ def finalize!
       # @api private
       def call(dir)
         registration_config = Configuration.new
+
         yield(registration_config) if block_given?
+
         components(dir).each do |component|
-          next if !component.auto_register? || registration_config.exclude.(component)
+          next if (
+            registered?(component) ||
```
jodosha, line 43: @timriley Should these inline conditions be extracted in one or more private methods to let the intent to emerge?

timriley: @jodosha Fair call :) Addressed in https://github.com/dry-rb/dry-system/pull/155/commits/515f76e6c09e3f5c0bcd1622f3beb3517701575c

lib/dry/system/auto_registrar.rb
```diff
@@ -26,37 +27,40 @@ def initialize(container)
...
-          .map { |file_name| [file_name, file_options(file_name)] }
-          .map { |file_name, options| component(relative_path(dir, file_name), **options) }
-          .reject { |component| registered?(component.identifier) }
+        files(dir).map { |file_name|
+          Component.new(
+            relative_path(dir, file_name),
+            file_path: file_name,
+            **MagicCommentsParser.(file_name)
+          )
+        }
```
timriley, line 59: By calling `Component.new` here we no longer initialize the component through `Container.component`, which is doing the extra work of trying to locate the component again across all the configured component_dirs. That's not necessary at all here, since we've already crawled a directory and have a known file path for each component! This change alone may have a measurable positive impact on the auto-registration phase for large projects!

lib/dry/system/booter.rb
```diff
@@ -54,6 +63,13 @@ def finalize!
         freeze
       end
 
+      # @!method finalized?
+      #   Returns true if the booter has been finalized
+      #
+      #   @return [Boolean]
+      #   @api private
+      alias_method :finalized?, :frozen?
```
timriley, line 71: Added this so we could use it inside `find_component`, since it much more naturally expresses the state of the `Booter` than `frozen?`, given that `finalize!` is the method that we call to finalize the booter.

lib/dry/system/booter.rb
```diff
@@ -115,12 +131,21 @@ def call(name_or_component)
...
 
-      # @api private
+      # Returns all boot files within the configured paths
+      #
+      # Searches for files in the order of the configured paths. In the case of multiple
+      # identically-named boot files within different paths, the file found first will be
+      # returned, and other matching files will be discarded.
+      #
+      # @return [Array<Pathname>]
+      # @api public
```
timriley, line 148: Added this comment as a clear reminder that we rely on this from other gems, namely dry-rails, effectively making it public API.

lib/dry/system/component.rb
```diff
@@ -47,49 +48,78 @@ class Component
...
       attr_reader :loader
 
+      # Returns a component with a namespace and path provided from a file found within
+      # the given component dirs. If no file is found, a component is returned without
+      # these attributes.
+      #
+      # @return [Dry::System::Component]
       # @api private
-      def self.new(*args, &block)
-        cache.fetch_or_store([*args, block].hash) do
```
timriley, line 52: @solnic I couldn't figure out why Component needed a cache, since every component should only get initialized once over the lifecycle of any given container, either as the auto-registrar runs, or as each component is lazily resolved. In the name of simplifying things while I worked, I removed the cache, but I'd love to know if there's any reason you think it needs to remain in place.

timriley:
@solnic @flash-gordon @jodosha This is now ready for review, complete with a thorough PR description!

timriley:
As a before/after comparison, here's how [`.load_component`](https://github.com/dry-rb/dry-system/blob/4c9c28c07f4369061dd6fc3d7f3263a67d7a8ae4/lib/dry/system/container.rb#L623-L643) and [`.load_local_component`](https://github.com/dry-rb/dry-system/blob/4c9c28c07f4369061dd6fc3d7f3263a67d7a8ae4/lib/dry/system/container.rb#L673-L689) used to look:

```ruby
def load_component(key, &block)
  return self if registered?(key)

  component(key).tap do |component|
    if component.bootable?
      booter.start(component)
    else
      root_key = component.root_key

      if (root_bootable = component(root_key)).bootable?
        booter.start(root_bootable)
      elsif importer.key?(root_key)
        load_imported_component(component.namespaced(root_key))
      end

      load_local_component(component, &block) unless registered?(key)
    end
  end

  self
end

def load_local_component(component, default_namespace_fallback = false, &block)
  if booter.bootable?(component) || component.file_exists?(component_paths)
    booter.boot_dependency(component) unless finalized?

    require_component(component) do
    end
  elsif !default_namespace_fallback
    load_local_component(component.prepend(config.default_namespace), true, &block)
  elsif manual_registrar.file_exists?(component)
    manual_registrar.(component)
  elsif block_given?
    yield
  else
    raise ComponentLoadError, component
  end
end
```

And here's how they look after this PR:

```ruby
def load_component(key)
  return self if registered?(key)

  component = component(key)

  if component.bootable?
    booter.start(component)
    return self
  end

  booter.boot_dependency(component)
  return self if registered?(key)

  if component.file_exists?
    load_local_component(component)
  elsif manual_registrar.file_exists?(component)
    manual_registrar.(component)
  elsif importer.key?(component.root_key)
    load_imported_component(component.namespaced(component.root_key))
  end

  self
end

private

def load_local_component(component)
  if component.auto_register?
  end
end
```

Hopefully this helps make it clear why I'm so excited about this PR!

lib/dry/system/container.rb
```diff
@@ -667,28 +633,41 @@ def inherited(klass)
           super
         end
 
-        private
+        protected
```
timriley, line 636: Look, it's a rare sighting of `protected` in its native habitat! I'm using this here because in `.load_imported_component` we need to call `.load_component` on the container that we're importing, which will be another `Dry::System::Container` instance, aka the exact reason `protected` exists :)

solnic:
> sidenote: I've followed this pattern in so many places now I think dry-configurable should support it out of the box

@timriley this must be added prior 1.0.0. dry-validation does this too and I suspect rom-rb's core config could benefit from this too.

lib/dry/system/component.rb
solnic: @timriley well, I don't remember 🤷🏻 it's possible that in the past it was needed and then it wasn't...and I just missed that fact.

solnic (approved):
@timriley thanks for this very detailed description, it made much more sense for me to carefully read it and understand the changes rather than to examine the diff. Since what you did here, conceptually, makes perfect sense to me, AND it resulted in simplified implementation which at the same time makes the library much more powerful, I have nothing to complain about 😆 FWIW - I've read the diff and nothing stands out as potentially problematic. I reckon seeing it work in real-world apps will be a much better verification process, it's a huge change after all.

This is *clearly* a huge milestone and I honestly didn't expect that the lib will be so greatly improved prior 1.0.0, so thank you for this huge effort, really ❤️ 

One thing I'll probably experiment with would be a backward-compatibility shim so that previous methods (that you removed) could still work. This should make it simpler to upgrade, but please tell me if this is a stupid idea.

I will also upgrade dry-rails!

Tim, seriously, again, this is an epic refactoring, I'm so happy. You're taking dry-system to the next level. Thank you! 🚀 🎉 🙇🏻

jodosha (approved):
@timriley This is impressive.

I was not familiar with the internals of the library, but I can tell that the implementation looks much better now.

By reading your detailed PR description and the integration specs, I've fully grasped the idea, and I'm very excited about it! Thanks, this is huge for `dry-system`! 🙌

lib/dry/system/config/component_dirs.rb
```diff
@@ -0,0 +1,52 @@
+require "concurrent/map"
+require "dry/configurable"
+require "set"
```
flash-gordon, line 3: doesn't seem to be used

flash-gordon (approved):
Great work! I left a few non-essential comments.
It also should fix a bug I came across yesterday about booting dependencies in the presence of a default namespace. I'll test it after merging this PR and send a spec.

lib/dry/system/component.rb
```diff
@@ -47,49 +48,78 @@ class Component
...
-          name, options = args
-          options = DEFAULT_OPTIONS.merge(options || EMPTY_HASH)
+      def self.locate(identifier, component_dirs, **options)
```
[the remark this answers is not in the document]
timriley: Good suggestion. I've just pushed https://github.com/dry-rb/dry-system/pull/155/commits/be928c3059aa10a60b460f1a0f9158db89a9a259 – is that the kind of thing you were meaning? Are there areas where you think we could take it further?

timriley:
@flash-gordon Thanks so much for the detailed feedback! I'm very happy to receive this; big changes like this are well worth fine-tuning, especially considering our attention is likely to turn elsewhere after this is done.

I've incorporated all your direct suggestions and force-pushed a new history. Have also left you a question about how best to avoid the object allocations through the keyword param splats.

lib/dry/system/component.rb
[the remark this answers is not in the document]
timriley: Also relevant is https://github.com/dry-rb/dry-system/pull/157/commits/199f37034eb00d13aad929296585464108450015 in my follow-up PR, where I continued to try and keep the options as whole hashes instead of splatted keyword args in some additional methods I added.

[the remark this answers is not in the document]
flash-gordon: yeah, like that. There's not much I can say about keywords in general, you just need to remember that `**varname` allocates a hash (inevitably). If a code path is hot you should think if this can be avoided. Also, mb ruby optimizes cases where you capture all keywords with `**options` and pass them untouched down with `**options`. But for all other cases allocation cannot be avoided. And actually it's 💯 the same for `*args`. Because of that we have [crazt code](https://github.com/dry-rb/dry-logic/blob/2e27997d0fdab90b3941e3ef71d93cf9ae62d591/lib/dry/logic/rule/interface.rb) in dry-logic.

PR #158, opened 2021-02-16
Introduce Dry::System::Identifier, refactor component building, and support component dirs with mixed namespaces

timriley:
This restores previous behavior that was lost during the introduction of the "component dir" concept. This means it's now possible to have a component dir configured with a default namespace, but also have components _outside_ that namespace still accessible via the container.

The crux of the change to reenable this behavior is in `ComponentDir#component_for_path`. And we have a new test to ensure it works as required: `spec/integration/container/auto_registration/mixed_namespaces_spec.rb`.

As part of this work, I also continued my dry-system refactoring journey, because this is more or less my life now 😆.

I removed the class-level factory/"builder" methods from `Component` (i.e. `.locate` and `.new_from_component_dir`) and moved their logic to `ComponentDir#component_for_identifier` and `#component_for_path`. Now all the methods for finding components within a dir are housed within the object that already represents that dir. It's a much better arrangement given how many aspects of the `ComponentDir` need to be accessed as part of finding these components. It also leaves `Component` itself much more focused.

With `#component_for_identifier` and `#component_for_path` together, the way that namespaced/non-namespaced components are handled across both loading strategies is much clearer, and it will be much easier to make sure we're providing consistent behavior (which is important since these methods are used for lazy-loading and finalised containers respectively, and it's important that these behave consistently!)

Along with this, I removed the confusing `Component.extract_identifier` and `.remove_namespace_from_name` methods that were called as part of initialising components. They're not needed anymore in this new arrangement, because the identifiers passed to components are already in the right form, with no further manipulation required.

Speaking of identifiers, I noticed we had a lot of repetitious string manipulation of identifier strings spread across the codebase, so I introduced `Dry::Component::Identifier` as a single class to manage all identifiers. It holds the identifier string, the default namespace (as provided by the component dir) and the namespace separator string. It also provides the `#path` for a given identifier, which is also an important detail to get right, and to have work consistently, because we use that whenever we `require` components.

As a flow on from this, I updated all aspects of the codebase that use identifiers to make sure they work with instances of `Identifier`. It feels like a really good cleanup overall!

A nice side-effect of having the new `Identifier` class is that it provides a place for us to add richer querying behaviour around identifiers, which dry-system _users_ can actually use as part of their own container configuration rules. For example, Nikita shared this:

```ruby
config.component_dirs.add('lib') do |dir|
  dir.default_namespace = 'my_app'
  dir.auto_register = lambda do |component|
    !component.identifier.start_with?('entities') &&
      !component.identifier.start_with?('external.entities')
  end
end
```

See that `component.identifier.start_with?("entities")`? Isn't that just so _nice?_ Much better than the old way of having to do e.g. `component.identifier =~ /^entities\./`.

We could add a handful more query methods like this as public API, but I think we should do that in a follow-up PR.

Anyway, please take a look at this overall and let me know if you're good for me to merge this in! This should be the last thing we need to do before we can make the next release of dry-system.


lib/dry/system/identifier.rb
```diff
@@ -0,0 +1,168 @@
...
+    #
+    # Components are eventually registered in the container using plain string
+    # identifiers, available as the `identifier` or `key` attribute here. Additional
+    # methods are provided to make it easier to evaluate or manipulate these identifiers.
+    #
+    # @api public
+    class Identifier
+      include Dry::Equalizer(:identifier, :namespace, :separator)
+
+      # @!attribute [r] identifier
```
solnic, line 18: This macro is not needed, YARD detects attr_reader syntax so a standard method tags will work here

lib/dry/system/identifier.rb
```diff
@@ -0,0 +1,168 @@
...
+      #   @api public
+      alias_method :to_s, :identifier
+
+      # @!method to_str
+      #   Returns the identifier string
+      #
+      #   @return [String]
+      #   @see #identifier
+      #   @api public
+      alias_method :to_str, :to_s
```
solnic, line 62: Why is this needed?

solnic (approved):
Approved but with a couple of comments 🙂 👏🏻 🙇🏻 ❤️

lib/dry/system/identifier.rb
timriley: Ah cool, thanks. I just copied those from elsewhere in the code. I'll adjust!

lib/dry/system/identifier.rb
timriley: It felt like the right thing to do for this class. Identifiers are strings at their core, so making them implicitly convertible to a string felt like the right behaviour. Means people working with an identifier could do things like `"foo " + component.identifier` and it'd just work.

You're right that it's not _strictly_ necessary, however. I'd be happy to remove it if you think it's a step too far. What do you think?

solnic: @timriley if we want this to act as a string, we should have it as a decorator. IIRC `to_str` should be used when something quacks like a string and it can be coerced into a string when needed. Otherwise it'll be potentially surprising to the user that `start_with?` works but some other String methods do not.

timriley: Fair point, I can remove it :)

jodosha (approved):
@timriley Great work! 👏 

> As part of this work, I also continued my dry-system refactoring journey, because this is more or less my life now 😆.

Made me smiley and grateful. Thanks for that! 🙏

flash-gordon (approved):
I already deployed it 👍🚀🚀🚀

lib/dry/system/component.rb
```diff
@@ -7,7 +7,7 @@
 require "dry/system/loader"
 require "dry/system/errors"
 require "dry/system/constants"
-require "dry/system/magic_comments_parser"
+require_relative "identifier"
```
solnic, line 10: @timriley out of curiosity - why `require_relative` here?

timriley: @solnic It's because I'd like to move most of the requires for "neighbouring" files to require_relative (it's slightly faster and it also is a clear signal that a file is coming from nearby), but I didn't want to touch lines that were not part of this overall change. I'll come back for those in a follow-up PR sometime soon.

solnic: @timriley ahh, great - do you happen to know how faster this is? We could totally change it everywhere in a batch-change to speed things up (across all the repos).

lib/dry/system/component.rb
```diff
@@ -41,88 +37,29 @@ class Component
...
-      def self.remove_namespace_from_name(name, namespace)
-        match_value = name.match(/^(?<remove_namespace>#{namespace})(?<separator>\W)(?<identifier>.*)/)
+        identifier =
+          if identifier.is_a?(Identifier)
+            identifier
+          else
+            Identifier.new(
+              identifier,
+              namespace: options.delete(:namespace),
+              separator: options.delete(:separator)
```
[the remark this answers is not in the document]
timriley: @flash-gordon Thanks for pointing this out. I considered it when writing this, and I thought it would be no real ill consequences of "leaking" those spurious options to the initializer. But I think you're right that this should be handled properly. I'll push up a fix for this before merging.

PR #162, opened 2021-03-10
Allow custom system-wide defaults to be configured for all component dirs

timriley:
This PR rounds out the work done in #155, #157, and #158, and makes it possible to configure default values for all component dir settings directly on the container's `config.component_dirs`. For example:

```ruby
config.component_dirs.loader = MyCustomLoader
config.component_dirs.default_namespace = "my_app"

# This dir will have the above values applied as defaults
config.component_dirs.add "lib"
```

This will be particularly useful e.g. for users configuring dry-system to work with an class autoloader like Zeitwerk, which would require the following defaults to be applied:

```ruby
config.component_dirs.loader = Dry::System::Loader::Autoloading
config.component_dirs.add_to_load_path = false
```

Having to configure these manually for every added component dir would be a suboptimal configuration experience. Further, the existence of these default settings will make it possible for us to create a plugin to enable a simple `use :zeitwerk` one-liner in the future.

It should be noted that the order of configuring the defaults and adding component dirs does not matter. The defaults will be applied to added component dirs regardless of whether they have been configured before or after the dirs are added.

Along with this, this PR fleshes out the API documentation for `Dry::System::Config::ComponentDirs` and `Dry::System::Config::ComponentDir`.

**This introduces one more BREAKING CHANGE** for the upcoming release: the top-level `loader` setting has been removed, with a system-wide loader now possible to configure via `component_dirs.loader`.

FYI, I plan to squash all commits here when merging.

### DIFF
--- a/lib/dry/system/config/component_dirs.rb
+++ b/lib/dry/system/config/component_dirs.rb
@@ -10,9 +10,10 @@ module Dry
       class ComponentDirs
         include Dry::Configurable
 
-        # Settings from ComponentDir can be configured here to apply all added dirs as
-        # defaults
-        @_settings = ComponentDir._settings.dup
+        # Settings from ComponentDir are configured here as defaults for all added dirs
+        ComponentDir._settings.each do |setting|
+          _settings << setting.dup
+        end
 
         # @!group Settings
 
--- a/lib/dry/system/config/component_dirs.rb
+++ b/lib/dry/system/config/component_dirs.rb
@@ -44,27 +44,20 @@ module Dry
 
         private
 
-        # Apply global default settings to a component dir. This is run every time the
-        # dirs are accessed, so this must be idempotent
+        # Apply default settings to a component dir. This is run every time the dirs are
+        # accessed, so this must be idempotent
         #
         # @return [void]
         def apply_defaults_to_dir(dir)
-          # Copy the existing config so we don't lose it after applying defaults
-          dir_config = dir.config.dup
+          dir.config.values.each do |key, value|
+            # For each component dir setting, if the value is still the setting's default,
+            # but the value configured _here_ is different, then apply it, since it must
+            # have been explicitly configured as a default for all component dirs
+            setting_default = Undefined.coalesce(dir.class._settings[key].default, nil)
+            system_default_value = public_send(key)
 
-          # Apply the defaults
-          config.values.each do |key, val|
-            dir.public_send(:"#{key}=", val)
-          end
-
-          # Reapply the dir's own config over the defaults, but only the values that are
-          # different from the setting defaults. This ensures we don't overwrite globals
-          # with meaningful, user-configured values.
-          dir_config.values.each do |key, val|
-            default_value = Undefined.coalesce(dir.class._settings[key].default, nil)
-
-            if val != default_value
-              dir.public_send(:"#{key}=", val)
+            if value == setting_default && system_default_value != value
+              dir.public_send(:"#{key}=", system_default_value)
             end
           end
         end
--- a/lib/dry/system/config/component_dirs.rb
+++ b/lib/dry/system/config/component_dirs.rb
@@ -2,7 +2,6 @@ require "concurrent/map"
 require "dry/configurable"
 require "dry/system/constants"
 require "dry/system/errors"
-require "dry/system/loader"
 require_relative "component_dir"
 
 module Dry
@@ -11,11 +10,8 @@ module Dry
       class ComponentDirs
         include Dry::Configurable
 
-        setting :auto_register, true
-        setting :add_to_load_path, true
-        setting :default_namespace
-        setting :loader, Dry::System::Loader
-        setting :memoize, false
+        # Component dirs settings can be configured here to apply as defaults to all dirs
+        @_settings = ComponentDir._settings.dup
 
         def initialize
           @dirs = Concurrent::Map.new
--- a/lib/dry/system/config/component_dirs.rb
+++ b/lib/dry/system/config/component_dirs.rb
@@ -113,6 +113,7 @@ module Dry
           raise ComponentDirAlreadyAddedError, path if dirs.key?(path)
 
           dirs[path] = ComponentDir.new(path).tap do |dir|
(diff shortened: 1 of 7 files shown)


lib/dry/system/config/component_dirs.rb
```diff
@@ -1,49 +1,181 @@
...
+        # Returns true if a setting has been explicitly configured, and not returning just
+        # a default value.
+        #
+        # This is used to determine which settings should be applied to added component
+        # dirs as additional defaults.
+        #
+        # @api private
+        def configured?(key)
+          config._settings[key].input_defined?
+        end
```
timriley, line 167: I had to dig into some dry-system internals here, but it _vastly_ simplifies the logic of `apply_defaults_to_dir` above.

Previously I was having to inspect the current values for both the system-wide and individual dir settings, the compare them to the settings own intrinsic default value, and then try and work out from those whether or not to apply the system-wide default. But this would fall down in the case of having a system-wide default that you want to _opt out of_ by explicitly configuring your individual dir to use the default value again.

For example:

```ruby
component_dirs.default_namespace "my_app"

component_dirs.add do |dir|
  # Explicitly provide nil here, even though it is already the "natural" default for this
  # setting, to opt out of the system-wide default configured above
  dir.default_namespace = nil
end
```

IMO, this is a reasonable behavior to support, and without the `#configured?` methods I added to `ComponentDirs` and `ComponentDir`, it would not be possible to achieve this.

And really, the best case to make for this approach is that `#apply_defaults_to_dir` is just so easy to read! `if configured?(key) && !dir.configured?(key)` – simple!

timriley:
@solnic @flash-gordon @jodosha — this is done and ready for review! Finally, this new feature is _done_ done 😅

jodosha (approved):
@timriley Nice one! 💯 I assume this would simplify a lot the internals of Hanami 2. 👍

lib/dry/system/config/component_dirs.rb
solnic: > I had to dig into some dry-system internals here

@timriley you mean dry-configurable internals, right?

lib/dry/system/config/component_dirs.rb
```diff
@@ -1,49 +1,181 @@
...
 module Dry
   module System
     module Config
       class ComponentDirs
-        attr_reader :dirs
+        include Dry::Configurable
 
+        # Settings from ComponentDir can be configured here to apply all added dirs as
+        # defaults
+        @_settings = ComponentDir._settings.dup
```
solnic, line 15: @timriley why is this an ivar?

solnic (approved):
Approved but I've got a couple of minor comments 🙂

lib/dry/system/config/component_dirs.rb
timriley: This is a class-level ivar, and it's where dry-configurable stores its setting definitions. By assigning it here, it lets us copy over the settings that are already defined for each individual `ComponentDir` and allow them to be used as defaults.

Another option could've been to create a module that contains the `setting` definitions and mix it into both places, but with the way things are set up, it IMO felt better for `ComponentDir` to "own" those settings.

What do you think?

lib/dry/system/config/component_dirs.rb
timriley: @solnic Oops, yes, you're right, I did mean dry-configurable internals :)

lib/dry/system/config/component_dirs.rb
solnic: @timriley let's leave it like that but my understanding is that this is a missing feature in dry-configurable, right? If yes, then it's something to improve there so that we could simplify things here 🙂

timriley: Now that I'm looking at this again a month later(!), I definitely agree that it'd good to have this handled in a nicer way. Ideally the dry-rb gems wouldn't require deep knowledge of the internals of _other_ dry-rb gems to do their job.

The thing is, this _could_ be easily done in a much cleaner way, like I suggested above, where we have a module that we mix into both `Config::ComponentDir` and `Config::ComponentDirs` that defines the settings in each. 

However, if I put the API documentation for the setting in that module, it wouldn't make sense for both including contexts (where one is about configuring individual dirs, and the other about configuring the defaults across _all_ dirs). I think that was pretty much what pushed me in this direction, to have `Config::ComponentDir` "own" the canonical setting definitions, and `Config::ComponentDirs` to copy them over for use when configuring defaults. This also let me define the API documentation in each place, with appropriate wording for each context.

The other thing I had in mind when doing this was what I might need to do if e.g. individual `ComponentDir` instances had a setting that I _didn't_ want configurable globally. In that case, it'd be simple enough to adjust this part here to loop over `ComponentDir._settings` and use an allow/deny list to determine which settings to include/skip when applying here. I'm probably overthinking it, though.

Perhaps one way in of avoiding the nasty class-level ivar assignment here would be to go right ahead do something like I just suggested:

```ruby
ComponentDir._settings.each do |setting|
  _settings << setting.dup
end
```

This is effectively the same as what I'm achieving with `@_settings = ComponentDir._settings.dup` but it does have the benefit of not breaking encapsulation and dealing with the underling ivar directly.

I'm inclined to go ahead and do this. What do you think, @solnic?

timriley: If we wanted a nicer public API for importing a whole range of settings from another place, I'd advocate for something like `Settings#replace`, which we could use in this context like so:

```ruby
_settings.replace(ComponentDir._settings.dup)
```

`#replace` has precedence across a range of classes within the Ruby stdlib, and since `Settings` is an enumerable collection, having this method here wouldn't be surprising.

solnic: @timriley thanks for the insights! I love the idea of using `#replace`.

timriley: I've now updated this to use the following:

```ruby
ComponentDir._settings.each do |setting|
  _settings << setting.dup
end
```

Will file another issue about adding `Settings#replace` to dry-configurable :)

PR #173, opened 2021-04-27
Fix dependency_graph-plugin

wuarmin:
Hey,

i stumbled upon an issue while using the dependency_graph-plugin:
```txt
undefined method `to_sym' for #<Dry::System::Identifier:0x000056140520e7b0>
# ./lib/dry/system/plugins/dependency_graph.rb:38:in `register'
```
This changes would fix my issue. In addition, the event keys are now all symbols. That wasn't the case. The dependency_graph-spec has also been improved.

What do you think?

kind regards

### DIFF
--- a/lib/dry/system/plugins/dependency_graph.rb
+++ b/lib/dry/system/plugins/dependency_graph.rb
@@ -34,10 +34,10 @@ module Dry
         # @api private
         def register(key, contents = nil, options = {}, &block)
           super
-          sym_key = key.to_sym
-          unless config.ignored_dependencies.include?(sym_key)
+          dependency_key = key.to_s
+          unless config.ignored_dependencies.include?(dependency_key)
             self[:notifications].instrument(
-              :registered_dependency, key: sym_key, class: self[key].class
+              :registered_dependency, key: dependency_key, class: self[dependency_key].class
             )
           end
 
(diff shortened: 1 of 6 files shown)


solnic (changes requested):
Thanks for tackling this. Unfortunately what should happen is ensuring those are always strings. Working with symbols like ":'foo.bar.baz'" is not nice 🙂

lib/dry/system/plugins/dependency_graph.rb
```diff
@@ -34,10 +34,10 @@ def self.dependencies
         # @api private
         def register(key, contents = nil, options = {}, &block)
           super
-
-          unless config.ignored_dependencies.include?(key.to_sym)
```
wuarmin, line 38: Alright, so `ignored_dependencies` should also be an array of strings?

solnic: Yes definitely! dry-system allows strings as identifiers in auto-injection modules BUT they are all coerced to strings internally, and that's how **dry-container** stores items, so this has to be consistent

wuarmin: Ok, that makes things clear to me. I'll fix this today.

solnic (approved):

solnic:
Thanks! This should go out soon as 0.19.1

wuarmin:
Thanks for the merge!

P.S.: If you're interested, I could fix the dry-system-examples (https://github.com/dry-rb/dry-system/tree/master/examples). They are outdated. What do you think?

solnic:
@wuarmin yes that would be very helpful!

PR #175, opened 2021-05-02
Check for actual registered string keys in AutoRegistrar

timriley:
This prevents "There is already an item registered with the key" Dry::Container::Error exceptions from being raised when finalizing the container after already having lazily loaded a few components.

This PR includes a new test case that would fail with the previous behaviour.

### DIFF
--- a/lib/dry/system/auto_registrar.rb
+++ b/lib/dry/system/auto_registrar.rb
@@ -32,7 +32,7 @@ module Dry
         components(component_dir).each do |component|
           next unless register_component?(component)
 
-          container.register(component.identifier, memoize: component.memoize?) { component.instance }
+          container.register(component.key, memoize: component.memoize?) { component.instance }
         end
       end
 
@@ -51,7 +51,7 @@ module Dry
       end
 
       def register_component?(component)
-        !container.registered?(component) && component.auto_register?
+        !container.registered?(component.key) && component.auto_register?
       end
     end
   end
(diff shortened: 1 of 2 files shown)


lib/dry/system/auto_registrar.rb
```diff
@@ -32,7 +32,7 @@ def call(component_dir)
         components(component_dir).each do |component|
           next unless register_component?(component)
 
-          container.register(component.identifier, memoize: component.memoize?) { component.instance }
+          container.register(component.key, memoize: component.memoize?) { component.instance }
```
solnic, line 35: Why not rely on duck-typing here? dry-container does `to_s` IIRC. This would be more future-proof.

solnic (approved):
✅ but I left one question (no biggie) '

lib/dry/system/auto_registrar.rb
timriley: @solnic I considered it, yep, but I wondered whether `#to_s` actually made sense for the _whole component_ object as we pass it around within dry-system. The `Component` actually has a range of different responsibilities and it didn't feel right to reduce it down to its identifier only via `#to_s`.

timriley: Happy to reconsider this and tweak in future, though!

flash-gordon:
This is actually a bug and it hasn't been released yet. Given master uses the new API from dry-configurable which is not released too things get messy. We should probably cut a branch from 0.19 and make a bug-fix release. I stumbled upon this problem while upgrading my largest app (I expected it wouldn't be smooth :).

flash-gordon:
Released in https://github.com/dry-rb/dry-system/releases/tag/v0.19.1

PR #178, opened 2021-05-05
Update finalize! signature in Container::Stubs

mpokrywka:
Currently Container.enable_stubs! causes Container.finalize! signature to change:
method no longer accepts "freeze" keyword argument. This causes consumers that
rely on this api to crash, i.e. dry-rails railtie

### DIFF
--- a/lib/dry/system/stubs.rb
+++ b/lib/dry/system/stubs.rb
@@ -7,7 +7,7 @@ module Dry
     class Container
       # @api private
       module Stubs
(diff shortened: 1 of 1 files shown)


solnic:
Thanks, I merged this even though it seems like we have no specs for this tricky thingie. I'll be working on a better solution before 1.0.0 where this entire stubs module will be most likely gone. This should be based on a simple configuration setting.

PR #181, opened 2021-06-22
Introduce component dir namespaces

timriley:
This change introduces support for **multiple, ordered, independent namespaces that a user can configure for any component dir,** which brings maximal flexibility to dry-system's code loading.

tl;dr

```ruby
class AdminContainer < Dry::System::Container
  configure do |config|
    config.root = __dir__

    config.component_dirs.add "lib" do |dir|
      dir.namespaces.add "admin", key: nil
    end
  end
end
```

## Background

To properly tell the story of namespaces, it's helpful to revisit the recent history of code loading in dry-system.

#### Pre-0.19.0

Before dry-system 0.19.0, you could configure a container to load code from multiple places by assigning multiple `auto_register` values:

```ruby
  configure do |config|
    config.root = __dir__
    config.auto_register = ["lib", "another_dir"] # paths relative to root
  end
end
```

You could also imperatively auto-register a directory by calling `auto_register!` directly (which allowed for additional code loading configuration to be applied):

```ruby
  configure do |config|
    config.root = __dir__
    config.auto_register = ["lib"]
  end

  auto_register! "another_dir" do |config|
    config.exclude do |component|
        # return true to exclude component from auto-registration
    end
  end
end
```

The trouble with this combination of approaches is that:

- They did not bring feature parity (e.g. registration rules could not be provided when providing `auto_register` dirs as config)
- They are not consistent in terms of their impact on container finalization (e.g. using `auto_register!` would eagerly crawl and register its provided dir, rather than supporting lazy loading for non-finalised containers)

Along with this, dry-system at the time offered just one way of providing special treatment of a particular "namespace" of components, via its `default_namespace` setting, which could be used like so:

```ruby
class AdminContainer < Dry::System::Container
  configure do |config|
    config.root = __dir__
    config.auto_register = ["lib"]

    # Any source files in `lib/admin/` will have the `"admin"` portion of their key removed
    config.default_namespace = "admin"
  end
end
```

This handled the common case of a container chiefly managing a single namespace of components, e.g. an admin container managing a directory of files largely in `lib/admin`, while ensuring that the redundant `"admin."` prefix wasn't part of every component's container key. 

#### 0.19.0

dry-system 0.19.0 unified the handling of directories of source files by introducing **component dirs as container configuration**. This provided a single facility for both adding and configuring directories from which to load components, with consistency behaviour for both lazy loading and finalized containers.

```ruby
  configure do |config|
    config.root = __dir__

    # Defaults for all component dirs can be configured separately
    config.component_dirs.auto_register = true # default is already true

    # Component dirs can be added and configured independently
    config.component_dirs.add "lib" do |dir|
      dir.add_to_load_path = true # defaults to true
      dir.default_namespace = "my_app"
    end

    # All component dir settings are optional. Component dirs relying on default
    # settings can be added like so:
    config.component_dirs.add "custom_components"
  end
end
```

Notably, a `default_namespace` could be specified for each individual component dir. This would allow for different container key namespace rules within each distinct component dir, but no ability to provide multiple rules _within_ a specific dir.

Another constraint remained at this point, too, in that dry-system expected the Ruby class constant inside each source file to adhere to the convention of the constant name matching the file's path within its component dir, e.g. `foo/bar/baz.rb` must provide `Foo::Bar::Baz`.

### The present challenge: supporting elided component dirs in Hanami 2 slices

This brings us to the present challenge, and the reason for this change in dry-system: supporting “elided” component dirs in Hanami 2’s slices.

By “elided,” I mean taking the only structure that dry-system 0.19.0 would support:

```
# Defines Admin::Foo

slices/admin/lib/admin/foo.rb 
```

And removing that second redundant `admin` in the path:

```
# Still defines Admin::Foo!

slices/admin/lib/foo.rb 
```

The motivation for this change in Hanami's app structure is simply to remove yet another rough edge that may pose as a hurdle to users.

To support this, we need to address one of those last remaining areas of inflexibility in dry-system: its expectations of the class constant defined within each component's source file (and as a bonus step, also address the related inflexibility around component key namespaces).

## Introducing component dir namespaces

In addition to dry-system already supporting multiple component dirs for a given container, we now support multiple **namespaces** within any given component dir.

```ruby
class AdminContainer < Dry::System::Container
  configure do |config|
    config.root = __dir__

    config.component_dirs.add "lib" do |dir|
      dir.namespaces.add "admin", key: nil
    end
  end
end
```

The container above is a simple minimal of the `AdminContainer` example provided earlier in our history, where we had to rely on a `config.default_namespace = "admin"`.

This change deprecates `default_namespace` and replaces it with this support for multiple `namespaces` within each component dir.

### Namespaces consist of a path, a key namespace, and a const namespace

A namespace consists of three elements:

- The **path** within the component dir to which it should apply the namespace. A namespace can only be added once for any given path.
- A **`key:`** namespace, which determines the leading part of the key used to register each component in the container. It can be:
  - Omitted, in which case it defaults to the value of `path`
  - A string, which will become the leading part of the registered keys
  - `nil`, which will make the registered keys top-level, with no additional leading parts
- A **`const:`** namespace, which is the Ruby namespace expected to contain the class constants defined within each component's source file.

  This value is provided as an "underscored" string, and will be run through the container's inflector to be converted in to a real constant (e.g. `"foo_bar/baz"` will become `FooBar::Baz`). Accordingly, `const:` can be:
  - Omitted, in which case it defaults to the value of `path`
  - A string, which will be constantized to the expected constant namespace per the rules above
  - `nil`, to indicate the class constants will be in the top-level constant namespace

To illustrate these options, some examples:

<br>

_An equivalent of the previous `default_namespace` setting:_

```ruby
namespaces.add "admin", key: nil
```

- `admin/foo.rb` is expected to define `Admin::Foo`, will be registered as `"foo"`
- `admin/bar/baz.rb` is expected to define `Admin::Bar::Baz`, will be registered as `"bar.baz"`

<br>

_Adding a distinct `const:` namespace:_

```ruby
namespaces.add "admin", key: nil, const: "super/admin"
```

- `admin/foo.rb` is expected to define `Super::Admin::Foo`, will be registered as `"foo"`
- `admin/bar/baz.rb` is expected to define `Super::Admin::Bar::Baz`, will be registered as `"bar.baz"`

<br>

_Omitting `key:`, with keys then naturally keeping an `"admin"` prefix:_

```ruby
namespaces.add "admin", const: "super/admin"
```

- `admin/foo.rb` is expected to define `Super::Admin::Foo`, will be registered as `"admin.foo"`
- `admin/bar/baz.rb` is expected to define `Super::Admin::Bar::Baz`, will be registered as `"admin.bar.baz"`

### Each component dir may have multiple namespaces

The leading example above showed a single namespace within a component dir, but a component dir may have any number of namespaces:

```ruby
class AdminContainer < Dry::System::Container
  configure do |config|
    config.root = __dir__

    config.component_dirs.add "lib" do |dir|
      dir.namespaces.add "admin/system_adapters", key: nil, const: nil
      dir.namespaces.add "admin", key: nil
      dir.namespaces.add "elsewhere", key: "stuff.and.things"
    end
  end
end
```

When the container loads its components, namespaces are searched and evaluated in order of definition. So for the example above:

1. Files within `lib/admin/system_adapters/` will have the `key: nil, const: nil` namespace rules applied
2. All other files in `lib/admin/` will have the `key: nil` namespace rules applied
3. Files in `lib/elsewhere/` will have the `key: "stuff.and.things"` namespace rules applied

### A root namespace is implicitly appended to a component dir's configured namespaces

To ensure that all the the files within a component dir remain loadable, a "root namespace" is implicitly appended to the list of configured namespaces on a component dir.

A root namespace, as the name implies, encompasses all files in the component dir. In the example above, the root namespace would be used when loading files _not_ in the `admin/` or `elsewhere/` paths.

The default root namespace is effectively the following:

```ruby
namespaces.add nil, key: nil, const: nil
```

It has `nil` path (the root of the component dir), a `nil` leading key namespace (all keys will be determined based on the full file path from the root of the dir), and a `nil` const namespace (implying that the root of the component dir will hold top-level constants).

These assumptions tend to hold true for typically organised projects, and they ensure that the component dirs can load code usefully even when no namespaces are configured at all.

### The root namespace may be explicitly configured

There may be cases where you want different namespace rules to apply when loading components from the root of the component dir. To support this, you can configure the root namespace explicitly via `namespaces.root`.

In this example, files in `lib/` are all expected to provide class constants in the `Admin` namespace:

```ruby
class AdminContainer < Dry::System::Container
  configure do |config|
    config.root = __dir__

    config.component_dirs.add "lib" do |dir|
      dir.namespaces.root const: "admin"
    end
  end
end
```

Root namespaces can be configured alongside other namespaces, and the same ordering preferences apply to them (i.e. for a given component file, the first matching namespace will “win” and its rules will be used).

### Namespaces may be configured as defaults for all component dirs

Like all other component dirs, namespaces may be added directly onto the container class' `config.component_dirs` to serve as defaults for all added component dirs:

```ruby
class AdminContainer < Dry::System::Container
  configure do |config|
    config.root = __dir__

    # This namespace will be configured for all component dirs
    config.component_dirs.namespaces.root const: "admin"

    config.component_dirs.add "lib"
  end
end
```

### The `default_namespace` setting has been deprecated

With the introduction of component dir namespaces, the `default_namespace` setting is no longer needed and has been deprecated.

For now, assigning a `default_namespace` still works, and will print a deprecation notice. So the following:

```ruby
config.default_namespace = "admin"
```

Will become this namespace under the hood:

```ruby
config.namespaces.add "admin", key: nil
```

## Implementation details

This PR isn't small. Sorry about that. To help you work through it, here are some notes on the implementation to guide you.

_n.b. to keep the class names from being over long, I've omitted `Dry::System` in most cases below_

### Introduced new `namespaces` setting

Firstly, `Config::ComponentDir` has a new `namespaces` setting that defaults to a `Config::Namespaces.new`. This is a mutable object (so users can `#add` namespaces as shown above), so the setting is `cloneable: true`, just like we already do for `Container`'s `component_dirs` setting.

To ensure we properly include `namespaces` in our feature of allowing component dir default settings to be set directly on `Container`'s `config.component_dirs`, we had to provide a special case for it in `Config::ComponentDir#configured?`, due to it being a mutable object assigned as the setting _default_ rather than the user explicitly assigning it as a whole object during configuration.

### Tidied up default component dir setting handling

While I was taking care of this, I made the handling of the defaults in `Config::ComponentDirs` more _compositional_, by instantiating a new `Config::ComponentDir` object and using that object to hold the defaults, rather than making wholesale copies of all the `ComponentDir` settings directly into `ComponentDirs` itself.

This change also allowed me to keep the special-casing of `namespaces` in `ComponentDir#configured?` in just one place, rather than having to spread it across both `ComponentDir` _and_ `ComponentDirs`.

### Deprecated `default_namespace` setting

The `default_namespace` setting on `Config::ComponentDir` is still in place, but its behaviour is mapped to adding and configuring a single namespace with `key: nil`. It also prints a deprecation warning. We'll want to remove this before we get to 1.0.

### Introduced `Config::Namespaces` and `Config::Namespace`.

These are the new classes that provide the configuration hooks for namespaces. They're quite straightforward.

One thing worth noting is that `Config::Namespaces#to_a` is the place where we add the implicit trailing root namespace if it's not otherwise configured. This worked quite neatly given the two sole consumers of `Config::Namespaces` (`ComponentDir#component_for_identifier` and `ComponentDir#each_component`) both just call `Config::Namespaces#each` only.

### `ComponentDir` component loading now runs through namespaces

With the namespaces in place, `ComponentDir` now iterates through the configured namespaces as the first step in each of its code loading paths: `#component_for_identifier` (for lazy loading) and `#each_component` (for finalized containers).

`ComponentDir#each_component` is new, replacing the previous `#component_for_path` method that was used for finalized container loading. The old implementation presumed an already-determined file path for the component (which `AutoRegistrar` used to determine, because it was doing the filesystem crawling), but this no longer makes sense now that we needed a place to ensure the configured namespaces were respected and handled consistently for all cases of building components (both lazy loading and finalizing).

With the new arrangement, all filesystem crawling done during finalizing happens inside `ComponentDir`, and the `AutoRegistrar` becomes much simpler orchestration class: it has very little internal logic, relying mostly on the components yielded from `ComponentDir#each_component`.

### `Component` initialized with a `namespace`, and all components can now be considered “loadable”

The components returned from `ComponentDir` (which are instances of `Component`) are now initialized with the namespace that was used to find them.

The `file_path` attribute has also been removed from `Component`, since we have enough information from the namespace to infer what we need about its file.

With `file_path` removed, we do need another way to signal whether a `Component` instance is directly “loadable.” 
`Container#load_component` wants to know this, because if the component is loadable, `#load_component` can go ahead and register it on the container directly, but if not, it will search for the component indirectly, via a manual registrar file or an imported container.

To support this, we now have a `Component#loadable?` method, which for `Component` will _always_ return `true`.

This is because inside `Container#find_component` (renamed in this PR from `Container#component`), we now instantiate an `IndirectComponent` when a component is requested but cannot be found via a directly loadable file inside any of the container's component dirs. `IndirectComponent#loadable?` returns `false`, as you'd expect. An `IndirectComponent` is one that will need loading via either a manual registration file of imported container.

This separation allows `Component` to carry much more "confident" code (fewer nil checks for `file_path`, etc.), and makes it clearer that we have two distinct kinds of components at play.

### Reordered bootable component loading to further simplify `Component`

To even further simplify `Component`, we've adjusted `Container#load_component` to search for a bootable component directly, early in the method body, and return early in the case of a bootable component being found.

This allows us to remove `Component#bootable?` and for the remainder of `#load_component` to be confident it's dealing with an ordinary `Component` (or `IndirectComponent`), not a bootable component.

### `Component#path` replaced with `#require_path` and (new!) `#const_path`

Now that we know every `Component` represents a loadable file, and now that each component has the namespace used to find it, we can present these new useful methods for loading the component:

- `#require_path`, which returns a path that can be passed to `Kernel#require`
- `#const_path`, which returns a stringified representation of the component's const namespace, able to be passed to the inflector's `#constantize`

Both of these are then used from within `Loader` to load an actual component instance.

`Loader.require!` also becomes a little simpler, because it no longer needs to check whether the component's file exists – it can now be certain that if a component has been passed to it, then it represents something loadable (and therefore require-able).

### Greatly simplified `Identifier`

Now that each `Component` is initialized with its namespace, this no longer needs to be passed to `Identifier`. `Identifier` also happened to hold the logic behind `Component#path`, which has also been removed.

With these two changes, `Identifier` becomes much, much simpler: it's now simply the combination of a container key along with the configured separator. This gives it much greater clarity of purpose.

We still keep a couple of its useful query methods it already had: `#root_key` and `#start_with?`. We've removed `#path`, since that is now handled much more appropriately by the new path-related methods inside `Component`.

We've also provided a clearer, more flexible replacement for the previous `Identifier#dequalified` method, which is now  `#namespaced`. The new `#namespaced` implementation is much more flexible, because it supports the addition, removal, and _replacement_ of arbitrary leading segments of the identifier's key. We use the latter two cases in other parts of the dry-system codebase, and the new name and method signature makes the intended result much more revealing.

### Clarified “Identifier” / “key” naming across the codebase

Another minor change I made to `Identifier` was to rename its internal attribute from `#identifier` to `#key`, mostly because I didn't like the length and repetition. This made me realise we find better consistency on these terms across the codebase. Along these lines, `key` is now used whenever we refer to a container key represented by a plain string, and `identifier` is only used when referring to an `Identifier` instance.

### Simplified `Component` initialization by requiring an `Identifier`

On a related note, `Component`'s initialization code was over-complex, because it supported receiving identifiers as both plain strings as well as `Identifier` instances.

It turns out that everywhere in our production code, we're already passing `Identifier` instances, and it was only the unit tests that passed plain strings. I updated the unit tests and deleted all that gross extra code!

### Introduced spec helpers for generating fixture directories

When I created the initial integration tests to drive the early development of this feature, I followed our existing practice of creating files in `spec/fixtures/` to be loaded by our test `Container` instances. However, as I expanded the tests to cover more variations, I became frustrated with these files acting as “mystery guests” inside the spec files – they were far-removed and their content and purpose was not at all self-evident.

To address these, I introduced `spec/support/tmp_directory.rb`, with helper methods for generating temporary fixture directories for test cases and writing files into it. Alongside this, I refactored our `$LOAD_PATH` and `$LOADED_FEATURES` cleaning code and moved it into `spec/support/loaded_constants_cleaning.rb`.

Together, this allowed me to write much clearer integration tests for the namespaces. You can see these in action at `spec/integration/container/auto_registration/component_dir_namespaces/`. Here's an example:

```ruby
    context "distinct constant namespace" do
      let(:component_dir_config) {
        -> dir {
          dir.namespaces.add "ns", const: "my_namespace"
        }
      }

      before :context do
        @dir = make_tmp_directory

        with_directory(@dir) do
          write "lib/ns/component.rb", <<~RUBY
            module MyNamespace
              class Component
              end
            end
          RUBY

          write "lib/ns/nested/component.rb", <<~RUBY
            module MyNamespace
              module Nested
                class Component
                end
              end
            end
          RUBY
        end
      end

      let(:cleanable_modules) { super() + %i[MyNamespace] }

      context "lazy loading" do
        it "resolves the component as an instance of a class in the given constant namespace" do
          expect(container["ns.component"]).to be_an_instance_of MyNamespace::Component
          expect(container["ns.nested.component"]).to be_an_instance_of MyNamespace::Nested::Component
        end
      end

      context "finalized" do
        before do
          container.finalize!
        end

        it "resolves the component as an instance of a class in the given constant namespace" do
          expect(container["ns.component"]).to be_an_instance_of MyNamespace::Component
          expect(container["ns.nested.component"]).to be_an_instance_of MyNamespace::Nested::Component
        end
      end
    end
```

No more flicking between the spec file and some vary-far-away fixture files to figure out what's going on. No more having to try and figure out how many test cases you can wring from a single fixture directory before it becomes too widely used and too hard to maintain.

Now, when you want to see what's going on, _you just scroll up!_ And when you want to introduce subtle variations for your examples, you can create dedicated fixture files just for those cases. So much nicer! I wish I did this years ago!

## Breaking changes

There are only a couple of small breaking changes in this PR:

- `ComponentDirNotFoundError` is no longer raised during finalization
- `Component#path` has been removed (replaced by `#require_path` and `#const_path`), but this is unlikely to impact users, unless they were providing their own custom `Loader` replacements


timriley:
Going to merge and release this now :)

Issue #166, opened 2021-04-11
Rename settings DSL `key` to `setting`


timriley:

timriley:
Fixed in #202.

PR #182, opened 2021-07-05
Renamed setting dsl method key to setting

choallin:
@timriley mentioned in the [ticket 166](https://github.com/dry-rb/dry-system/issues/166) that the setting DSL method key should be renamed to setting, which is done in this pr

### DIFF
--- a/lib/dry/system/settings.rb
+++ b/lib/dry/system/settings.rb
@@ -29,9 +29,18 @@ module Dry
           end
         end
 
-        def key(name, type)
+        def setting(name, type)
+          if ::Kernel::__callee__ == :key
+            ::Dry::Core::Deprecations.announce(
+              "the name of this method will be renamed to :settings",
+              "Please refactor your call at #{::Kernel::caller[0]}",
+              tag: "dry-system",
+              uplevel: 2
+            )
+          end
           schema[name] = type
         end
+        alias :key :setting
       end
 
       class Configuration < Dry::Struct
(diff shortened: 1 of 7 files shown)


lib/dry/system/settings.rb
```diff
@@ -29,9 +29,18 @@ def call
...
+            ::Dry::Core::Deprecations.announce(
+              "the name of this method will be renamed to :settings",
+              "Please refactor your call at #{::Kernel::caller[0]}",
+              tag: "dry-system",
+              uplevel: 2
+            )
+          end
           schema[name] = type
         end
+        alias :key :setting
```
solnic, line 43: There's actually a `deprecate` macro in `Deprecations` extension that you could use here https://github.com/dry-rb/dry-core/blob/master/lib/dry/core/deprecations.rb#L152-L157

solnic (commented):
Thanks for addressing this. I left one small suggestion.

lib/dry/system/settings.rb
choallin: Thanks for your hint!
Since the DSL class is inheriting from BasicObject and is not including any Dry module I don't have access to the deprecate call. Can you point me to a case where there is a similar case so I can take a look on how to do this?

wuarmin: Check out https://github.com/dry-rb/dry-system/blob/176ead290bb5018bd8e6b5a69226eda51392b5ed/lib/dry/system/container.rb#L98

timriley:
Thanks for taking a run at this, @choallin! In the end this got addressed as part of a whole overhaul of the ~bootable components~ providers in #202, so this is fixed now 😄

PR #183, opened 2021-07-12
Fix settings component example code

rickenharp:
Since `Dry::System::Settings::DSL` is a `BasicObject`, it doesn’t include `Kernel`, so `#proc` is not defined.
To get this example working, call `Kernel.proc` directly

### DIFF
--- a/docsite/source/settings.html.md
+++ b/docsite/source/settings.html.md
@@ -16,7 +16,7 @@ require "path/to/dry/types/file"
 Application.boot(:settings, from: :system) do
   settings do
     key :database_url, Types::String.constrained(filled: true)
-    key :logger_level, Types::Symbol.constructor(Kernel.proc { |value| value.to_s.downcase.to_sym })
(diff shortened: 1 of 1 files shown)


docsite/source/settings.html.md
```diff
@@ -16,7 +16,7 @@ require "path/to/dry/types/file"
 Application.boot(:settings, from: :system) do
   settings do
     key :database_url, Types::String.constrained(filled: true)
-    key :logger_level, Types::Symbol.constructor(proc { |value| value.to_s.downcase.to_sym })
+    key :logger_level, Types::Symbol.constructor(Kernel.proc { |value| value.to_s.downcase.to_sym })
```
solnic, line 19:
```suggestion
    key :logger_level, Types::Symbol.constructor { |value| value.to_s.downcase.to_sym }
```

rickenharp: Oh, that's even better!

timriley:
Thanks!

Issue #159, opened 2021-02-22
Disallow direct access to `config` until `configure` has been run


timriley:
Running `Dry::System::Container.configure`, passing a block, and configuring the container on the yielded object is the recommended way to configure a dry-system container, because we run certain hooks after `.configure` is run for the first time (such as adding configured component_dirs to Ruby's `$LOAD_PATH`).

It's also possible to configure a container by assigning values on it's `.config` directly. This is just how dry-configurable works. Doing this within the context of `Dry::System::Container`, however, may lead to surprising behavior, since it can't provide an opportunity for the after-configure hooks to run.

To improve this situation, we should either:

- Raise an error if `config` is access directly without `configure` having been run first
- Find some other ergonomic way for the user to signal that the after-configuration hooks should run

The former is much simpler than the latter.

solnic:
I believe we want option nr 1 :)

timriley:
I've reflected on this now, and I think it would be undesirable for us to forbid direct access to `config` at any time. This is how dry-configurable is _intended_ to work, and it seems non-ergonomic to do anything else.

Instead, I'm working to expand the API around the overall configuration _phase_ for `Dry::System::Container`. Work in progress is at #207.

PR #184, opened 2021-07-12
Forbid direct write access to the config object

choallin:
Direct access to config isn't allowed anymore [ticket 159](https://github.com/dry-rb/dry-system/issues/159)

Tests are all green, but this still feels a little bit of sluggish.

Also, right now we never get rid of the newly defined method_missing and therefore it is *never* possible to directly write a config.

Feedback is very much appreciated!

### DIFF
--- a/lib/dry/system/container.rb
+++ b/lib/dry/system/container.rb
@@ -135,7 +135,6 @@ module Dry
 
         def config
           if @hooks_initialized != true && finalized? != true && @configure_self != true
-            klass = super
             super.instance_eval do
               def method_missing(meth, *args)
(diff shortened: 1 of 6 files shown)


timriley:
Thanks so much for taking a run at this, @choallin! As your code certainly shows, this is a tricky problem to solve! After reflecting on it for a while, I've decided to sidestep this issue by expanding upon the tools we provide for managing the container configuration phase, and allowing our users both ways to configure their container, either via the `configure` block-based API, or using `config` directly. Work in progress is at #207, FYI, and I hope I can have this wrapped up in the next few days.

PR #185, opened 2021-07-13
Fix broken link to berg in index.html.md

Spone:


timriley:
Thanks for this @spone :) Sorry for the delay in responding to this! In merging #181 I've taken care of this dead link by removing that paragraph entirely. I don't think berg would serve as a good example of dry-system today anyway. Hopefully by the time we get closer to 1.0 we can have a couple of good examples inside the repository for users to refer to.

PR #195, opened 2021-12-01
Broader interfaces to Config::ComponentDirs and Config::Namespaces

timriley:
This PR broadens the interfaces of both the `Dry::System::Config::ComponentDirs` and `Dry::System::Config::Namespaces` collection classes to make them more useful to external consumers looking to programatically construct and inspect these configurations. Specifically, it supports the addition of `source_dirs` config to Hanami in https://github.com/hanami/hanami/pull/1133.

The updated API for `Config::ComponentDirs` includes:

- `#dir` (also aliased as `#[]`) has been added, for retrieving previously added dirs by their paths
- `#add` now accepts a single `path_or_dir` argument, supporting two distinct usages:
  - The existing usage, of providing a path string and then a block that is yielded a component dir object for that path, to be configured
  - A new usage, where a whole component dir object is already built and passed directly to `#add`
- `#delete` has been added, for deleting a previously added dir by its path
- `#length` has been added
- `#dirs` has been made protected, since it is the internal store for the dirs and may not always return dirs with the latest default values applied to them (`#to_a` and `#each` remain public and are preferred for this)

The updated API for `Config::Namespaces` includes:

- `#namespace` (also aliased as `#[]`) has been added, for retrieving previously added namespaces by their paths
- `#root` has been repurposed to returning a previously added root namespace (the previous usage is deprecated, but still supported for now)
- `#add_root` has been added, to take the place of the previous `#root` method, for adding a configured root namespace. This makes its purpose clearer and fits better alongside `#add`.
- `#delete` has been added, for deleting a previously added namespace by its path
- `#delete_root` has also been added, for deleting a previously added root namespace
- `#paths` has been added, returning an array of paths for the namespaces
- `#length` has been added

Along with this, API documentation has been lightly updated to make it clearer that these two collection config classes (along with their counterpart singular `Config::ComponentDir` and `Config::Namespace` classes) are public API.

I intend to make a new minor release of dry-system after this has merged, to allow the upcoming Hanami alpha to depend on it.

I'll also be squashing the commits in this PR when merging.

### DIFF
--- a/lib/dry/system/config/component_dirs.rb
+++ b/lib/dry/system/config/component_dirs.rb
@@ -189,6 +189,7 @@ module Dry
         def length
           dirs.length
         end
+        alias_method :size, :length
 
         # Returns the added component dirs, with default settings applied
         #
--- a/lib/dry/system/config/component_dirs.rb
+++ b/lib/dry/system/config/component_dirs.rb
@@ -109,7 +109,7 @@ module Dry
         #
         # @see ComponentDir
         def add(path_or_dir)
-          path, dir_to_add = prepare_to_add(path_or_dir)
+          path, dir_to_add = path_and_dir(path_or_dir)
 
           # TODO: is this worth even raising?
           raise ComponentDirAlreadyAddedError, path if dirs.key?(path)
@@ -117,16 +117,6 @@ module Dry
           dirs[path] = dir_to_add.tap do |dir|
             yield dir if block_given?
             apply_defaults_to_dir(dir)
-          end
-        end
-
-        private def prepare_to_add(path_or_dir)
-          if path_or_dir.is_a?(ComponentDir)
-            dir = path_or_dir
-            [dir.path, dir]
-          else
-            path = path_or_dir
-            [path, ComponentDir.new(path)]
           end
         end
 
@@ -167,6 +157,17 @@ module Dry
         end
 
         private
+
+        # TODO docs
+        def path_and_dir(path_or_dir)
+          if path_or_dir.is_a?(ComponentDir)
+            dir = path_or_dir
+            [dir.path, dir]
+          else
+            path = path_or_dir
+            [path, ComponentDir.new(path)]
+          end
+        end
 
         # Applies default settings to a component dir. This is run every time the dirs are
         # accessed to ensure defaults are applied regardless of when new component dirs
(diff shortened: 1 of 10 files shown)


lib/dry/system/config/component_dirs.rb
```diff
@@ -78,69 +80,170 @@ class ComponentDirs
...
+        end
+
+        # Returns the count of component dirs
+        #
+        # @return [Integer]
+        #
+        # @api public
+        def length
+          dirs.length
         end
```
solnic, line 191: can we have a size alias to make it consistent with `Array`? People typically use `size` anyway.

lib/dry/system/config/namespaces.rb
```diff
@@ -87,11 +110,54 @@ def add(path, key: path, const: path)
...
+
+        # Deletes the configured root namespace and returns the namespace
+        #
+        # If no root namespace was previously configured, returns nil
+        #
+        # @return [Namespace, nil]
+        #
+        # @api public
+        def delete_root
+          namespaces.delete(Namespace::ROOT_PATH)
```
solnic, line 138:
```suggestion
          delete(Namespace::ROOT_PATH)
```

lib/dry/system/config/namespaces.rb
```diff
@@ -87,11 +110,54 @@ def add(path, key: path, const: path)
...
+        def paths
+          namespaces.keys
+        end
+
+        # Returns the count of configured namespaces
+        #
+        # @return [Integer]
+        def length
+          namespaces.length
+        end
```
solnic, line 154:
```suggestion
        end
        alias_method :size, :length
```

lib/dry/system/config/component_dirs.rb
solnic, line 191:
```suggestion
        end
        alias_method :size, :length
```

solnic (approved):
Approved with some minor suggestions 🙂

lib/dry/system/config/component_dirs.rb
timriley: @solnic No probs! Just added these :) It's funny, I don't think I've ever personally used `#size` in preference to `#length`, I wonder how I wound up this way, haha

solnic: @timriley `size` is shorter and easier to type, I **never** use `length` 🙂

flash-gordon:
Just glanced over the well-written description 👍, only one question, why `size`/`length` at all? I mean if you need to check for emptiness it should be `empty?`, otherwise, how does it help if `.length` returns 4 or 5 for instance?

PR #197, opened 2021-12-12
Add zeitwerk plugin

### LINKED ISSUE #170: Add a `use :zeitwerk` plugin
I'd like to make zeitwerk + dry-system work together with a single line, so it should be possible to just throw this into your `Dry::System::Container` subclass:

```ruby
use :zeitwerk
```

This should do the necessary dry-system configuration:

```ruby
config.component_dirs.loader = Dry::System::Loader::Autoloading
config.component_dirs.add_to_load_path = false
```

And it could also perhaps require zeitwerk, set up the base zeitwerk loader, make it available on the container for further manual configuration, if required, and then `setup` the loader as part of the container boot lifecycle.

ianks:
_(Latest PR description from @timriley)_

This PR adds a plugin to make setting up zeitwerk easier. This makes using Zeitwerk with dry-system a one-liner:

```ruby
  use :zeitwerk # <=== here it is!

  configure do |config|
    config.root = Pathname(__dir__).join("..")

    config.component_dirs.add "lib" # <= will automatically be pushed to the zeitwerk loader
  end
end
```

The Zeitwerk plugin:

- Creates and configures a `Zeitwerk::Loader` instance and assigns it to `config.autoloader` (if will also respect a loader instance that you've assigned manually)
- Configures all component dirs to use our `Dry::System::Loader::Autoloading` loader, which loads classes by accessing their constant, instead of requiring them
- Configures all component dirs so they're not added to `$LOAD_PATH` by default
- Pushes all configured component dirs to the loader, including all their respective module namespaces (also defining those modules if they don't exist at the time)
- At container finalization time, tells the loader to `eager_load` if (a) we're also using the `:env` plugin and the env is `:production`, or (b) the user has specifically configured the plugin with `eager_load: true`

Fixes #170

To do:

- [x] Separate the pushing of dirs to the loader from configuring the loader, so if a user has assigned their own loader for any reason, we still push component dirs to it
- [x] Update original spec to generate source files rather than use static fixtures
- [x] Add `.dependencies` to the plugin module


ianks:
With regards to eager loading, I was thinking of having a setting on the plug-in for it, defaulting to true in production and false otherwise. Thoughts?

timriley:
This looks amazing @ianks, thanks for doing it! I hope to have some feedback for you in the next day or two 😀

ianks:
I added, what I think, is a wise strategy for eager-loading.

By default, the plugin will have zeitwerk eager load when using the `:env` plugin sets the environment to `:production`.

```ruby
  use :env, inferrer: -> { :production }
  use :zeitwerk # will eager load
end
```

However, you can change this behavior by passing `:eager_load` option to the plugin:

```ruby
  use :zeitwerk, eager_load: true
end
```

lib/dry/system/plugins/zeitwerk.rb
```diff
@@ -0,0 +1,103 @@
...
+        end
+
+        # Add component dirs to the zeitwerk loader
+        #
+        # @return [Zeitwerk::Loader]
+        #
+        # @api private
+        def push_component_dirs_to_loader(system, loader)
+          system.config.component_dirs.each do |dir|
+            raise ZeitwerkAddToLoadPathError, dir if dir.add_to_load_path
```
timriley, line 86: I wonder if we actually want to raise this error at all.

This nothing actually _wrong_ with having a directory in load path and also passing it with Zeitwerk. In that scenario, everything would still work as expected with Zeitwerk.

Given this, and given we already provide the default `add_to_load_path = false` up in `#configure_autoloading`, my feeling is that we should allow users to opt back into this behaviour if they so choose. At that point, they'd have to be doing it for a good reason and we should get out of their way.

What do you think about removing this error?

lib/dry/system/plugins/zeitwerk.rb
```diff
@@ -0,0 +1,103 @@
...
+            loader.push_dir(system.config.root.join(dir.path))
+          end
+
+          loader
+        end
+
+        # @api private
+        def eager_load?(system)
+          options.fetch(:eager_load) do
+            system.config.respond_to?(:env) && system.config.env == :production
```
timriley, line 97: This feels like a good default for eager_load and I like how you have it playing nicely with other plugins that we ship 👍🏼

lib/dry/system/plugins/zeitwerk.rb
```diff
@@ -0,0 +1,103 @@
...
+        #
+        # @api private
+        def setup_autoloader(system)
+          return system if system.registered?(:autoloader)
+
+          if system.config.autoloader
+            system.register(:autoloader, system.config.autoloader)
+          else
+            system.config.autoloader = build_zeitwerk_loader(system)
+          end
```
timriley, line 56: I'm a little confused by this... so if the user has provided their own autoloader instance via config, we register it in the container, but if we build our default one, we don't? What were you looking to achieve when you set this part up?

Would you actually want to register the autoloader in the container for both sides of that if/else clause? That said, I wonder if we actually want the autoloader in the container _at all_ — it feels too... _external_ to the application for us to want to have it readily accessible to all the application components themselves.

lib/dry/system/plugins/zeitwerk.rb
```diff
@@ -0,0 +1,103 @@
...
+
+        # Configure autoloading on the container
+        #
+        # @return [self]
+        #
+        # @api private
+        def configure_autoloading(system)
+          require "dry/system/loader/autoloading"
+
+          system.setting :autoloader, reader: true
```
timriley, line 35: I know you're following the pattern of the other plugins here, but now that I look at this, I feel like it'd make more sense to declare the setting as early as possible, i.e. in `#extended` rather than waiting for the before_configure hook. But for the sake of this PR, let's leave this consistent with the other plugins.

So: definitely not a blocker for this PR, but wanted to mention it given it became apparent to me over the course of this review.

@solnic — given you authored most of the existing plugins, is there any reason they declare their settings in before configure hooks, rather than in the standard module extended hook?

spec/support/zeitwerk_helpers.rb
```diff
@@ -0,0 +1,15 @@
+# frozen_string_literal: true
+
+module ZeitwerkHelpers
```
timriley, line 3: This is a helpful extraction, thank you for doing this!

spec/integration/container/plugins/zeitwerk_spec.rb
```diff
@@ -0,0 +1,86 @@
...
+      configure do |config|
+        config.root = SPEC_ROOT.join("fixtures/zeitwerk_eager").realpath
+
+        config.component_dirs.add "lib"
+      end
+    end
+
+    expect { $zeitwerk_eager_loaded }
+      .to output(/\$zeitwerk_eager_loaded' not initialized/)
+      .to_stderr
```
timriley, line 59: It doesn't feel super great to have our expectation be relying on warning output from Ruby because we're trying to access a non-initialised global var... would it be better to check for the presence of `:$zeitwerk_eager_loaded` within `global_variables` instead?

timriley (commented):
This looks like it's heading in a really good direction — awesome work @ianks! 🎉 

I've left a bunch of notes for you as inline comments. I hope it's not too overwhelming. Definitely happy to discuss anything in further detail if you need it 😄 

Regarding your questions in the PR description:

> What do you think of the name Container.autoloader? Maybe it should be Container.code_loader?

Good question. I definitely like the feel of `autoloader` to `code_loader`, and it does fit well with the name we're already using for the `Loader::Autoloading` class.

That said, I could see how it might be potentially confusing given we already have a `loader` that we use elsewhere in config.

Given this plugin is intended to work specifically with zeitwerk, I wonder if it's better to embrace that and just call it `zeitwerk_loader`? What do you think?

>  Do we need to configure namespaces with Zeitwerk? I'm assuming no

In the most simple configuration, i.e. component dirs with conventional Ruby source directory structures, then Zeitwerk namespaces wouldn't be required.

However, we already have Hanami's dry-system/zeitwerk integration showing us that there _is_ a need to configure these in a way that synchronises a namespace across both the dry-system component dir as well as the dir pushed to the zeitwerk loader.

See e.g. `lib/hanami/slice.rb` in https://github.com/hanami/hanami/pull/1123. In this case, we have a "compact" directory structure where everything in root of a component dir will already be part of a base Ruby module structure, so e.g. `lib/foo.rb` inside the container whose root is `slices/main/` will define `Main::Foo`. For Zeitwerk to properly load this, we need to push the `lib` dir to it along with the `Main` module as its `namespace:`.

So I think we should invest some time into seeing if there's to make this work nicely out of the box. Perhaps it could be as simple as looking for the `const:` option we pass to our component dir namespaces, and constantizing that in prep for passing to Zeitwerk when pushing the dir?

lib/dry/system/plugins/zeitwerk.rb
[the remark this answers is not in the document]
ianks: I’m not against removing it, I’m just wondering if it could cause strange behavior for those who are already adding this option. Likely, with zeitwerk, it would be a very rare use case yeah? If it is, it seems like it could be helpful to prevent potential footguns?

I’m OK either way, those are just my thoughts to why i raised there

lib/dry/system/plugins/zeitwerk.rb
```diff
@@ -0,0 +1,103 @@
...
+  module System
+    module Plugins
+      class Zeitwerk < Module
+        attr_reader :options
+
+        # @api private
+        def initialize(options)
+          @options = options
+          super()
+        end
```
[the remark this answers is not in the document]
ianks: Would prefer that! Wasnt sure on the dry.rb policy here but I'll happily switch 👍🏻

lib/dry/system/plugins/zeitwerk.rb
ianks: Fair enough. I just copy-pasta'd this from the logging plugin. Happy to make it just register on the container instead

lib/dry/system/plugins/zeitwerk.rb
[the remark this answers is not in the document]
ianks: actually I think I remember why I did it now. by using the options, we can lazily `require 'zeitwerk'` for those that don't use it. Another way around this would be to register the plugin and require zeitwerk in a registration block. Happy to change if you think that's better.

lib/dry/system/plugins/zeitwerk.rb
solnic: @timriley it was to ensure that plugin settings are available before configuration takes place, there were cases where doing it in the moment a plugin is required was too early or too late.

timriley:
After a conversation with @ianks  a couple of weeks ago, I'm taking on the final few bits of work on this PR. I'll do this over the rest of this week. Thanks for doing all the heavy lifting here, @ianks! ❤️

lib/dry/system/plugins/zeitwerk.rb
timriley: We're now declaring the settings directly on the container class in `extended`, like the rest of the plugins do after the changes in #207.

lib/dry/system/plugins/zeitwerk.rb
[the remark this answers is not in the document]
timriley: I took a stab at this and I realise there's enough _deferred_ default values based on the current state of the container that it doesn't make sense to use fixed kwargs and static defaults. Have left it as is :)

docsite/source/plugins.html.md
```diff
@@ -22,6 +22,77 @@ App[:logger]
...
+
+```ruby
+class App < Dry::System::Container
+  use :zeitwerk, eager_load: true
+end
+```
+
+### Debugging
+
+When you are developing your application, you can enable the plugin's debugging mode by passing `:debug` option to the plugin:
```
jcmfernandes, line 78: It would be great to see an explanation of what the debug mode means/does.

timriley: It's nothing fancy. Added some extra text to make it clearer :)

timriley:
@solnic @flash-gordon FYI, this feels ready to me now. Feel free to take a closer look, if you were wanting to! Will merge in the next couple days.

timriley:
@ianks Same for you, if you wanted to check over any of my adjustments :)

docsite/source/plugins.html.md
```diff
@@ -22,6 +22,77 @@ App[:logger]
...
+    config.component_dirs.add "lib"
+  end
+end
+```
+
+For a more in depth and runnable example, [see here](https://github.com/dry-rb/dry-system/tree/master/examples/zeitwerk).
+
+### Inflections
+
+The plugin passes the containeer's inflector to the Zeitwerk loader for resolving constants from file names. If Zeitwerk has resolving some constants, you can update the container's inflector like so:
```
cllns, line 50:
```suggestion
The plugin passes the container's inflector to the Zeitwerk loader for resolving constants from file names. If Zeitwerk has trouble resolving some constants, you can update the container's inflector like so:
```

docsite/source/plugins.html.md
```diff
@@ -22,6 +22,77 @@ App[:logger]
...
+  use :zeitwerk, debug: true
+end
+```
+
+### Advanced Configuration
+
+If you find you need to adjust Zeitwerk configuration, you can do so by accessing the `Zeitwerk::Loader` instance directly on the container, as `.autoloader`:
+
+```ruby
+# After you have configured the container
```
cllns, line 91:
```suggestion
# After you have configured the container but before you have finalized it
```
I think this is true?

lib/dry/system/container.rb
```diff
@@ -634,14 +636,35 @@ def importer
           @importer ||= config.importer.new(self)
         end
 
-        # Registers a callback hook to run afer container lifecycle events.
+        # Registers a callback hook to run before container lifecycle events.
```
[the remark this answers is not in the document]
timriley: Might just leave these as is, since in the docs they'll be clearly connected to the method name, and to me the bolding only really makes sense if the terms "before" and "after" are closer to each other than they actually are in this file.

PR #198, opened 2022-01-04
Add glossary to docs

timriley:


timriley:
@solnic Are you happy with these naming changes? If so, I'd like to start rolling them out as part of my 1.0 preparation work across the next two weeks.

timriley:
@jodosha Pinging you here too since this is a really critical foundation of Hanami, and I want to make sure you're comfortable with all the terms here too.

docsite/source/glossary.html.md
```diff
@@ -0,0 +1,71 @@
...
+
+**Source file:** A source file is a `.rb` Ruby source file inside a **component dir**, defining a single class with a name corresponding to the file. The container loads this file during **auto-registration** to register a matching **component**. (TODO: A source file may have **magic comments** to control its registration behavior)
+
+**Auto-registration:** Auto-registration one of the main reasons to use dry-system: it makes it easy to work with applications consisting of a large number of components. The container **auto-registers** components both when you **finalize** it, as well as when you **resolve** a component while the container is **lazy loading.**  When auto-registering, the container automatically registers a **component** for the class defined in each **source file** in each of its **component dirs**. The container matches the component to its source file based on the source file’s name and any **namespaces** you have configured in the component dirs.
```
solnic, line 27:
```suggestion
**Auto-registration:** Auto-registration is one of the main reasons to use dry-system: it makes it easy to work with applications consisting of a large number of components. The container **auto-registers** components both when you **finalize** it, as well as when you **resolve** a component while the container is **lazy loading.**  When auto-registering, the container automatically registers a **component** for the class defined in each **source file** in each of its **component dirs**. The container matches the component to its source file based on the source file’s name and any **namespaces** you have configured in the component dirs.
```

docsite/source/glossary.html.md
```diff
@@ -0,0 +1,71 @@
...
+
+**Memoized components**: By default, the **loader** will create a new **component instance** every time a component is **resolved**. However, you can also configure a component to be memoized (via **component dir** configuration or **magic comments**), in which case an instance is created the first time, and then reused for all subsequent resolutions.
+
+## Container lifecycle
+
+**Lazy loading:** Containers start in lazy loading mode, loading individual components just in time when you resolve them (as well as all the components for their dependencies, for components using **auto-injection**). This uses the same **auto-registration** process as when you **finalize** the container, ensuring your components are consistently available in both modes. Typically, you leave the container to lazy load when you want to optimize for fastest container load time, such as when running tests or your application in development mode.
```
solnic, line 47:
```suggestion
**Lazy loading:** Containers start in lazy loading mode, loading individual components just in time when you resolve them (as well as all the components for their dependencies, for components using **auto-injection**). This uses the same **auto-registration** process as when you **finalize** the container, ensuring your components are consistently available in both modes. Typically, you leave the container to lazy load when you want to optimize for fastest container load time, such as when running tests or in development environment.
```

solnic (approved):
This is great! 🎉 👏🏻

inouire:
Hi @timriley , just as an boserver here, I have always struggled a bit to understand some of the dry-* concepts in the past, and this document makes things clearer about dry-system

Something troubled me: finalize / finalization is used without being introduced, is this expected ?

And one question for my understanding: would lazy loading be used more in dev mode, and finalize used more on a prod app?

Thanks :pray:

docsite/source/glossary.html.md
timriley: I made an attempt at this change in https://github.com/dry-rb/dry-system/pull/198/commits/3d3c43b1efa48aed54dc2eea66f01fc176b8237b — seem OK to you?

solnic: @timriley perfect, TY ❤️

docsite/source/glossary.html.md
```diff
@@ -0,0 +1,71 @@
...
+
+**Manual registration**: You can manually register a **component** in the **container** via its `#register` method, passing the component's **key**, along with either a block that returns a new **instance** of the component (which will be called whenever the component is **resolved**), or a single object (to be returned as the **instance** value whenever the component is resolved).
```
cllns, line 31:
```suggestion
**Manual registration**: You can manually register a **component** in the **container** via its `#register` method, passing the component's **key**, along with either a block that returns a new **instance** of the component (which will be called whenever the component is **resolved**), or a single object (to be returned as the **instance** whenever the component is resolved).
```

docsite/source/glossary.html.md
```diff
@@ -0,0 +1,71 @@
...
+
+**Manifest registration:** You can create one or more registration manifest files, containing code that **manually registers** one or more components in the container. These files are searched during **auto-registration** to allow those components to be registered and **resolved** during both **finalization** and **lazy loading**.
```
cllns, line 33: Is this the only way, or the only practical way, or the preferred way, of doing manual registration?

docsite/source/glossary.html.md
```diff
@@ -0,0 +1,71 @@
...
+
+**Container imports:** You can specify other containers to import into your own. The components in the imported container will be registered in your own, with a key prefix you provide. The container will be imported in full as part of **finalization**, or individual components from the container will be imported as required when **lazy loading**.
```
cllns, line 35: What is a "key prefix"? Is it separated with an underscore or a period (though I think that's a namespace)? E.g. for a `car` container, if I had an `engine`, and the prefix was `car`, would it be `car_engine`?

docsite/source/glossary.html.md
```diff
@@ -0,0 +1,71 @@
...
+
+**Shutdown:** When you shut down a container, it runs the **stop hook** for each of its providers.
+
+## Container configuration
+
+**Container configuration**: You configure your container subclass via a class-level `config` object.
+
+**Container root**: This is the path for the root of your application. All **component dirs** are expected to be under this root. For typical applications, this will be the same as the your project directory. For a minimal container setup, you should configure both the container root and at least one **component dir**.
+
+**Component dir namespaces**: You can configure one or more namespaces within each **component dir**. Each namespace corresponds to a path within the dir, and sets configuration for loading components from the source files under that path: a **key namespace** (a prefix for the component key) and a **const namespace** (an expected Ruby constant to be used as the base namespace for the component's class constant).
```
cllns, line 59:
```suggestion
**Component dir namespaces**: You can configure one or more namespaces within each **component dir**. Each namespace corresponds to a path within the dir, and sets configuration for loading components from the source files under that path. This consists of both a **key namespace** (a prefix for the component key) and a **const namespace** (an expected Ruby constant to be used as the base namespace for the component's class constant).
```

docsite/source/glossary.html.md
```diff
@@ -0,0 +1,71 @@
...
+
+**Provider paths**: You can configure one or more paths to be searched when the container loads **providers**. By default, this is a single `"system/providers/"` path only. The container will search these paths in the order specified, with the first match for a given provider name being used to load that provider.
```
cllns, line 61:
```suggestion
**Provider paths**: You can configure one or more paths to be searched when the container loads **providers**. By default, this is a single `"system/providers/"` path within the **container root** only. The container will search these paths in the order specified, with the first match for a given provider name being used to load that provider.
```

docsite/source/glossary.html.md
```diff
@@ -0,0 +1,71 @@
...
+
+## Extensibility
+
+**Plugins**: You can activate additional functionality for your container by activating one or more plugins. Plugins can add additional settings and methods on your container, register their own components, and otherwise act in response to container lifecycle events (such as before/after initial configuration). When you activate a plugin, you may also pass additional options to tailor the plugin’s behavior.
+
+**Provider packs**: You can use prebuilt providers offering useful functionality by accessing **provider packs**. These are typically distributed via 3rd party Ruby gems. When you use a provider from a provider pack, you still create your own **provider file**, but instead of specifying the provider lifecycle yourself, you instead specify a provider from a pack, and supply any required configuration for the provider. You can also enhance the provider's own lifecycle by adding your own "before" and "after" hooks for the provider lifecycle events.
```
cllns, line 67:
```suggestion
**Provider packs**: You can use prebuilt providers offering useful functionality by accessing **provider packs**. These are typically distributed via 3rd party Ruby gems. When you use a provider from a provider pack, you still create your own **provider file**, but instead of specifying the provider lifecycle yourself, you instead specify a provider from a pack, and supply any required configuration for the provider. You can also enhance the provider's own lifecycle by adding your own `before` and `after` hooks for the provider lifecycle events.
```

docsite/source/glossary.html.md
```diff
@@ -0,0 +1,71 @@
...
+
+## Testing
+
+**Container stubs**: You can enable stubbing on the container to replace a particular component (via its key) for the sake of certain integration tests. You should hopefully only need this approach sparingly, since most classes designed to work well with the container (especially those using **auto-injection**) should be able to be tested in isolation using dependency injection, with particular dependencies replaced via test doubles passed to the class’ constructor.
```
cllns, line 71:
```suggestion
**Container stubs**: You can enable stubbing on the container to replace a particular component (via its key) for the sake of certain integration tests. You should hopefully only need this approach sparingly, since most classes designed to work well with the container (especially those using **auto-injection**) should be able to be tested in isolation using dependency injection, with particular dependencies replaced via test doubles passed to the class’ constructor as manual replacements.
```

cllns (commented):
Yes! This is great. I learned a ton. Made a number of suggestions. I had the gist of what `dry-system` did before, but now I feel like I understand it a lot better. I hope my perspective is useful!

Should we add an entry for **Manual replacements**, this is a pretty fundamental feature and it's mentioned once (twice after my suggestions)?

docsite/source/glossary.html.md
```diff
@@ -0,0 +1,71 @@
...
+
+**Provider:** A provider manages the lifecycle around configuring and registering one or more components (or setting global state if necessary) required for distinct parts of your application to work. When you register a provider (by creating a **provider file** in your **provider path**), you provide code to run for one or more of its **lifecycle hooks**, `prepare`, `start`, and `stop`. You typically create a provider when you need to register components with particular configuration (such as a client for a third party service, requiring API keys and other connection details) or when components are particularly heavyweight (such as a database persistence system). Every provider has a unique **name** that also corresponds to a **container key** **namespace**. Whenever any **component** in that namespace is **resolved** from the container, then the provider will be **started** (with its `prepare` and `start` hooks run in sequence). Providers can also be individually controlled via the container's `.prepare`, `.start`, and `.stop` methods.
```
cllns, line 29:
```suggestion
**Provider:** A provider manages the lifecycle around configuring and registering one or more components required for distinct parts of your application to work. When you register a provider (by creating a **provider file** in your **provider path**), you provide code to run for one or more of its **lifecycle hooks**: `prepare`, `start`, and `stop`. You typically create a provider when you need to register components with particular configuration (such as a client for a third party service, requiring API keys and other connection details, or setting global state if necessary) or when components are particularly heavyweight (such as a database persistence system). Every provider has a unique **name** that also corresponds to a **namespace** in a **container key**. Whenever any **component** in that namespace is **resolved** from the container, then the provider will be **started** (with its `prepare` and `start` hooks run in sequence). Providers can also be individually controlled via the container's `.prepare`, `.start`, and `.stop` methods.
```
GitHub isn't previewing the diff properly for me so just in case it's not clear:
1: Add a colon before the lifecycle hooks to show it's an exhaustive list
2. Move the 'setting global state' part to later, since it's not something we want to encourage but may be necessary.
3. `**container key** **namespace**` renders as the same as `**container key namespace**`. I'm not sure if the wording I used here, of `**namespace ** in a **container key**` is quite right either though.

docsite/source/glossary.html.md
```diff
@@ -0,0 +1,71 @@
...
+title: Glossary
+layout: gem-single
+name: dry-system
+---
+
+## The basics
+
+**Container:** The container is central to dry-system. When you use dry-system, your very first step is to create your own `Dry::System::Container` subclass, which will be the class you use to manage and access all of your application’s **components**.
+
+**Component key:** A component’s **key** is a string that uniquely identifies that component within the **container**. You can **resolve** a component from the **container** by passing its key to the container’s `.[]` method.
```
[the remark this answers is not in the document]
cllns: Should we also specify that a **namespace** for a **key**  looks like "foo.bar" here? We reference key namespaces but not that they're separated from the rest of the key with a period.

cllns (commented):
Tried to fix one of the diffs (highlighting deletions/additions) but it's not working, so listed out the changes.

jodosha (approved):
@timriley This is an epic work. 👏 LGTM.
Only one thing is puzzling me: the difference between Plugin and Provider Packs.

---

Next iteration for this documentation should be working on:

  * Formatting: split those long paragraphs to ease the read.
  * Linking: use anchors to link concepts within the same doc. Readers shouldn't hold all the information in their head, but be able to jump back and forth between information.
  * Code examples/snippers: how the reader applies what they just read?

solnic:
> Only one thing is puzzling me: the difference between Plugin and Provider Packs.

@jodosha a plugin *extends dry-system itself* to enable more functionality, a provider pack gives you components that you use within your app.

flash-gordon (approved):
Not only glossary is extremely helpful, this doc alone is enough to become a "confident user" :)

PR #199, opened 2022-01-04
Stop delegating methods to container in boot DSL

### LINKED ISSUE #164: Remove method_missing behavior from bootable component lifecycle DSL
Particularly these bits:

```ruby
      def method_missing(meth, *args, &block)
        if target.registered?(meth)
          target[meth]
        elsif container.key?(meth)
          container[meth]
```

There's too much magic in this behavior, and the concision it may add doesn't outweigh just how much harder it makes it to understand the behavior within the bootable component lifecycle blocks.

timriley:
There's too much magic in this behavior, and any concision it may add doesn't outweigh just how much harder it makes it to understand the behavior within these lifecycle blocks.

Fixes #164

### DIFF
--- a/lib/dry/system/lifecycle.rb
+++ b/lib/dry/system/lifecycle.rb
@@ -109,7 +109,6 @@ module Dry
 
       private
 
-      # @api private
       def trigger!(name, &block)
         if triggers.key?(name)
           triggers[name].(target)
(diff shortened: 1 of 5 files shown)


solnic (approved):

flash-gordon:
`include Kernel`?

flash-gordon:
or `< Object` even?

timriley:
In the end I decided to keep this behaviour in the #202 provider overhaul. Feels useful enough and we've made the code evaluation context much clearer thanks to the class-backed providers.

PR #200, opened 2022-01-05
Rename “bootable components” to “providers” (and the “init” lifecycle step to “prepare”)

timriley:
This PR renames "bootable component" (or "component" as it was confusingly termed in a few places) to "provider".

It includes a range of related changes in order for the end state to be coherent:

- Renamed "bootable component" (or "component" in some places) to "provider"
- Renamed the previous "provider" term to "source provider" - since it's job is to provide the _source_ for a concrete provider within a container
  - With the source providers, there's no longer two different abstractions (previously these were "providers" offering a collection of individual "components") — now it's just source providers _only_, with each provider having a "group" name as a way to bring them together and avoid naming conflicts when multiple groups of source providers are being used
- Renamed the `key:` param when using a source provider to `source:`
- Renamed the provider "init" lifecycle hook to "prepare"
- Clarified some of the terminology inside the `Provider` class itself:
  - The two possible blocks are passed as named arguments, rather than one of them as Ruby's implicit trailing `&block`
  - The first is called `lifecycle_block`, which is the block that should hold the definition of the actual provider lifecycle (this is provided by the block that users write themselves for their own providers, or by the source provider when using one of those)
  - The second is called `refinement_block`, which is the block that is passed only when using a source provider. In this case the job of this block is to _refine_ the lifecycle already established by the source provider
  - Renamed the `#finalize` method to `#apply`, since its job is to apply the registrations from the lifecycle container into the target container. In addition, it's actually run more than once per provider (after both the "prepare" and "start" hooks), so "finalize" as a term didn't make sense anymore anyway.
- Renamed the `Lifecycle` class to `ProviderLifecycle`
- Renamed the `bootable_dirs` container setting to `provider_dirs`, with a default value of `"system/providers"`
- While I was here, removed the completely unused `system_dir` container setting

One thing to note that's different from my glossary PR is that we no longer really need the idea of "provider packs" — instead, 3rd parties will just register groups of source providers. There was no particular functionality that needed to be added via a "pack" abstraction. The only purpose of group name when registering source providers is to make sure there's no unwanted naming clashes, that's all, and it's serving its purpose well just as a single arg when registering the source provider.

Along with these name changes, I did a bunch of minor refactors where possible just to improve general clarity and separation of concerns.

There's a bunch of churn here to get these name changes in place, but the benefits should hopefully be clear. Chiefly, throughout the codebase (and in what we communicate to our users), we have just one name for this concept now, instead of two ("provider" instead of a mix of "bootable component" and "component", depending on which bit of code you were looking at).

There's benefit in the code too. For example, in the `Booter` we have the `boot_dependency(component)` method. What the `component` here was unclear previously, because we used the term "component" for multiple things. However, with this change now, this is the only place inside `Booter` that actually deals with a `component` (because everything else is now a `provider`), which means it's much clearer that we're dealing with a `Dry::System::Component` instance in this case.

In reviewing this PR, it might actually be helpful to step through it **commit by commit**, because then you can see my thinking in action (and I tried to keep each step fairly discrete).

TODO:

- [x] Add backwards compatibility as much as possible


timriley:
This is effectively ready for review, @flash-gordon and @solnic — the only thing missing is API doc updates, plus any backwards compatibility and deprecation concerns. I'm going to move onto that tomorrow, but if you're interested in taking a look before then, please do!

timriley:
I have one more API doc to update (the chunky one above `Container.register_provider`), but otherwise this is now fully done! 

Even includes full backwards compat of and deprecation notices for the historical API, along with supporting tests.

@solnic @flash-gordon @jodosha This is ready for if/when you want to take a look. I'm going to be basing all future work on top of this branch, so if you're happy with the naming changes, be good for us to get this merged 😄

solnic (approved):
👏🏻 👏🏻 👏🏻

timriley:
Thanks for checking this out, @solnic 😄 

There's a bunch more tidying/clarifications I'd like to do, but I think we've bitten off enough for this particular PR! I'll merge it now and move onto more targeted efforts next.

timriley:
FWIW, it'll be a week or more until all this is ready for a proper release, so please feel free to leave any additional thoughts on this PR in the meanwhile.

PR #202, opened 2022-01-11
Provider and booter overhaul, including class-backed provider sources

timriley:
After renaming "bootable component" to "provider" in #200, this PR completely overhauls the Provider and Booter structure and capabilities.

The biggest change is that it backs the `register_provider` block-based DSL with _real classes_ implementing each provider's behavior, via `Dry::System::Provider::Source`. This "Provider Source" is defined by the block-based DSL in two ways:

- When you're defining a standalone provider inside your application, a new subclass of `Provider::Source` is created, then the block passed to `Dry::System::Container.register_provider` is used to define the body of that class. Specifically, it will define real _instance methods_ for each of the lifecycle steps: `#prepare`, `#start`, and `#stop`.
- When you're defining a provider that uses an external provider source, then the provider source (via `Dry::System.register_provider_source`) will have already created and defined a subclass of `Provider::Source` with instance methods for its specific behavior, and then the block passed to `.register_provider` is then evaluated in the context of the _instance_ of that source, allowing you to provide config (via the built-in `Provider::Source#configure`) or define lifecycle step callbacks (via the built-in `Provider::Source#before` and `#after`).

The block DSL evaluation context has become much simpler in this change, too. There's just two clear contexts, which are hopefully quite clear now that we have class-backed provider sources:

- When defining a source, `self` is the `Provider::Source` subclass
- When defining a source, inside the body of the `prepare`, `start`, `stop` blocks, `self` is the `Provider::Source` subclass _instance_
- Likewise _using_ an external source, `self` is the `Provider::Source` subclass _instance_

For the latter two blocks, we no longer need block arguments to access the provider container or the target container. These are available as methods that are already available on `self` in each of the block evaluation contexts (`#provider_container`/`#container` and `#target_container`/`#target`). For the initial block, access to the target container is intentionally limited because any taget container access should be done inside the latter two block contexts anyway. This should make the providers easier to reason about because there's now just one way to access the provider/target containers.

As the final step of provider setup, a `Dry::System::Provider` is initialized, and is provided its specific `Provider::Source` subclass. `Provider` takes care of all the standard behavior for each Provider, such as setting up the provider container, which it passes to the Source (among other things) when it initializes it. `Provider` also takes care of all the provider lifecycle, making sure that e.g. lifecycle steps aren't run when they don't need to be, and that we merge components from the provider container into the target container after each lifecycle step.

With `Provider` taking full responsibility for the standard Provider behavior and lifecycle, this has allowed the (former) `Booter` to become much simpler: it's now focused on the loading of the provider files themselves, and then registering, storing, and providing access to the Provider instances. The rest of its behavior is just simple delegations to those Provider instances. I've renamed this to `ProviderRegistrar` to reflect its new behavior and fit better with the overall set of names we have now.

---

One really nice benefit of the new class-based provider sources is that we can now _create our own source classes directly,_ rather than having to use the DSL only. This can be really helpful for complex providers:

```ruby
module Test
  class LoggerProvider < Dry::System::Provider::Source
    def prepare
      require "logger"
    end

    def start
      logger = complex_custom_logic_to_build_the_logger(config)
      register(:logger, logger)
    end

    private

    def complex_custom_logic_to_build_the_logger(config)
      # ...
    end
  end
end

Test::Container.register_provider(:logger, source: Test::LoggerProvider)
```

Better yet, class-based provider sources allow us to _share state_ between the lifecycle steps! Previously the only way we could do this was by having to register components in the provider container. For example:

```ruby
module Test
  class LoggerProvider < Dry::System::Provider::Source
    def prepare
      @internal_thing = "something we may want to set up initially but we still need inside start"
    end

    def start
      # Look, I can access @internal_thing here, because this is just a PORO!
    end

    private

    def complex_custom_logic_to_build_the_logger(config)
      # ...
    end
  end
end

Test::Container.register_provider(:logger, source: Test::LoggerProvider)
```

But it doesn't stop there, because this also takes external provider sources to the next level. With class-based sources we can now have our external sources _provide their own methods_ to be used when those sources are used when registering a provider. I've used this approach to straighten out our first-party `settings` provider source, which previously hijacked the provider `settings` object (which was intended only for providers to configure their sources) to load settings from env. This always twisted my brain whenever I had to think about it. Now, it's much more straightforward. See the source class:

```ruby
module Dry
  module System
    module ProviderSources
      module Settings
        class Source < Dry::System::Provider::Source
          def prepare
            require "dry/system/provider_sources/settings/settings"
          end

          def start
            config = settings.load(target.root, target.config.env).config
            register(:settings, config)
          end

          def settings(&block)
            # Save the block and evaluate it lazily to allow a provider with this source
            # to `require` any necessary files for the block to evaluate correctly (e.g.
            # requiring an app-specific types module for setting constructors)
            if block
              @settings_block = block
            elsif @settings_class
              @settings_class
            elsif @settings_block
              @settings_class = Class.new(ProviderSources::Settings::Settings, &@settings_block)
            end
          end
        end
      end
    end
  end
end

Dry::System.register_provider_source(
  :settings,
  group: :dry_system,
  source: Dry::System::ProviderSources::Settings::Source
)
```

You'll see here that it defines its _own_ `#settings` instance method, which can then be used when the provider source is used to register a provider:

```ruby
register_provider(:settings, from: :dry_system) do
  before(:prepare) do
    target_container.require_from_root "types"
  end

  settings do
    setting :database_url, constructor: SettingsTest::Types::String.constrained(filled: true)
    setting :session_secret, constructor: SettingsTest::Types::String.constrained(filled: true)
  end
end
```

This all just works now because of those updated block-based DSL evaluation contexts, with the `register_provider` context above being the _instance_ of the provider source, which lets us provide instance-level settings in a natural way, while still having those settings available to the lifecycle methods of the provider source (because they too are instance methods of the same object), which lets us take objects provided from the `register_provider` block and use it from within the source behavior.

I could imagine this being really helpful for complex provider sources. For example, if we wanted to provide one for rom-rb and we wanted to provide lots of targeted hooks for users to supply their own config or behaviour, in our provider source class we could set up as many instance methods as we want for this.

Finally, some other notable changes made in this PR:

- The default provider source class-level `settings` and instance-level `config` are now powered by dry-configurable rather than us building a custom `Dry::Struct`. This feels much more natural and more in-keeping with our overall ecosystem.
- Renamed the previous "source provider" concept to "provider source", since it fits much more naturally with the class structures and naming schemes we have in place in the code in this PR
- Renamed "dry/system/components" to "dry/system/provider_sources" (and adjusted the class namespacing to match)
- Renamed the `group` for our provider sources to `:dry_system` (from `:system`) in order to set a healthier precedent for third party use (i.e. use your own gem name for your group to ensure no conflicts).

These changes should all be **fully backwards compatible**, with a suite of deprecation shims and warnings in place.


jodosha (approved):
@timriley Excellent work. I wasn't able to follow the massive changes, but the PR description looks exciting!

lib/dry/system/container.rb
```diff
@@ -240,9 +240,17 @@ def import(other)
         # @return [self]
         #
         # @api public
-        def register_provider(name, namespace: nil, from: nil, source: nil, &block)
+        def register_provider(name, namespace: nil, from: nil, source: nil, &block) # rubocop:disable Metrics/PerceivedComplexity
```
solnic, line 243:
```suggestion
        # rubocop:disable Metrics/PerceivedComplexity
        def register_provider(name, namespace: nil, from: nil, source: nil, &block)
```

lib/dry/system/provider/source.rb
```diff
@@ -0,0 +1,320 @@
...
+require_relative "source_dsl"
+
+module Dry
+  module System
+    class Provider
+      # A provider's source provides the specific behavior for a given provider to serve
+      # its purpose.
+      #
+      # Sources should be subclasses of `Dry::System::Source::Provider`, with instance
+      # methods for each lifecycle step providering their behavior: {#prepare}, {#start},
```
solnic, line 15:
```suggestion
      # methods for each lifecycle step providing their behavior: {#prepare}, {#start},
```

lib/dry/system/provider_registrar.rb
```diff
@@ -0,0 +1,200 @@
...
+      # @!method finalized?
+      #   Returns true if the booter has been finalized
+      #
+      #   @return [Boolean]
+      #   @api private
+      alias_method :finalized?, :frozen?
+
+      # @api private
+      def shutdown
+        providers.values.each do |provider|
```
solnic, line 135:
```suggestion
        providers.each_value do |provider|
```

lib/dry/system/provider_registrar.rb
```diff
@@ -0,0 +1,200 @@
...
+          provider.prepare
+          self
+        end
+      end
+
+      # @api private
+      def start(name_or_provider)
+        with_provider(name_or_provider) do |provider|
+          provider.start
+          self
```
solnic, line 152: Shouldn't `self` be below the block?

lib/dry/system/provider_sources/settings/file_parser.rb
```diff
@@ -0,0 +1,59 @@
+# frozen_string_literal: true
+
+# TODO: Just update this to try dotenv and be done with it
```
solnic, line 3: 👍🏻 💯

lib/dry/system/provider_sources/settings/settings.rb
```diff
@@ -0,0 +1,63 @@
...
+          end
+
+          private
+
+          def setting_errors(errors)
+            errors.sort_by { |k, _| k }.map { |key, error| "#{key}: #{error}" }
+          end
+        end
+
+        class Settings
```
solnic, line 29: Rename to `Config`?

solnic (approved):
Approved but I left some comments/suggestions

lib/dry/system/provider_registrar.rb
timriley: This was how I found it, but I agree it'd be clearer at the very bottom!

timriley:
Thanks for the feedback so far, folks 😄

@solnic, I applied all your suggestions!

I'll merge this one tomorrow. Happy for continued feedback post-merge, too, of course!

lib/dry/system/provider_sources/settings/config.rb
```diff
@@ -0,0 +1,77 @@
...
+              env_data = load_files(root, env)
+
+              errors = {}
+
+              new.tap do |settings_obj|
+                settings.to_a.each do |setting_name|
+                  value = ENV.fetch(setting_name.to_s.upcase) { env_data[setting_name.to_s.upcase] }
+
+                  begin
+                    settings_obj.config.send(:"#{setting_name}=", value) if value
```
flash-gordon, line 41: `send` vs `public_send`?

lib/dry/system/provider_sources/settings/config.rb
```diff
@@ -0,0 +1,77 @@
...
+            errors.sort_by { |k, _| k }.map { |key, error| "#{key}: #{error}" }
+          end
+        end
+
+        class Config
+          class << self
+            def load(root, env)
+              env_data = load_files(root, env)
+
+              errors = {}
```
flash-gordon, line 34: you can move it to the `tap`'s block I think, or use `errors = ...each_with_object({}) do |setting_name, errors|`

spec/fixtures/test/system/providers/client.rb
```diff
@@ -1,9 +1,11 @@
 # frozen_string_literal: true
 
 Test::Container.register_provider(:client) do
-  use :logger
+  start do
+    use :logger
 
-  Client = Struct.new(:logger)
+    Client = Struct.new(:logger)
```
flash-gordon, line 7: that's a leaking constant btw

flash-gordon: mb namespace it within `Test`?

flash-gordon (approved):

lib/dry/system/provider.rb
```diff
@@ -257,19 +209,39 @@ def lf_container
...
-          STR
+          raise ArgumentError,
+                "+namespace:+ must be true, string or symbol: #{namespace.inspect} given."
         end
       end
 
-      # Set config object
+      # @api private
+      def run_step(step_name)
+        return self if statuses.include?(step_name)
```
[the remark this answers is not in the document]
timriley: Yeah, you're right, I was trying something out here, documenting the return as `[void]` when I wanted to signal that the caller shouldn't really be intending to capture or do anything useful with the return value, but given this raised a flag to you as a code reviewer, I realise I should probably just make the API documentation faithfully represent the implemented reality.

I'm changing these back to `[self]` where appropriate.

lib/dry/system/provider.rb
```diff
@@ -31,222 +37,168 @@ module System
...
+      #
+      # @return [Dry::System::Provider::Source]
+      #
+      # @api private
+      attr_reader :source
 
-          event = :prepare
-        end
+      # @api private
+      def initialize(name:, namespace: nil, target_container:, source_class: nil, &block) # rubocop:disable Style/KeywordParametersOrder
```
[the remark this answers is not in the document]
timriley: Oohh, great pick up, thank you @flash-gordon! In a previous iteration I had this method building the source class from the provided &block, and I didn't remember to update the param when I removed that behaviour. Glad you found this!

lib/dry/system/provider.rb
```diff
@@ -257,19 +209,39 @@ def lf_container
...
       #
       # @api private
-      def configure!
-        @config = settings.new(Components::Config.new(&@config_block)) if settings
+      def apply
+        provider_container.each do |key, item|
+          target_container.register(key, item) unless target_container.registered?(key)
+        end
+
+        self
```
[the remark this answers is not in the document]
timriley: Left this as `self` per my previous comment about the api docs 😄

lib/dry/system/provider/source.rb
```diff
@@ -0,0 +1,327 @@
...
+      class Source
+        class << self
+          # Returns a new Dry::System::Provider::Source subclass with its behavior supplied by the
+          # given block, which is evaluated using Dry::System::Provider::SourceDSL.
+          #
+          # @see Dry::System::Provider::SourceDSL
+          #
+          # @api private
+          def for(name:, group: nil, target_container:, &block) # rubocop:disable Style/KeywordParametersOrder
+            Class.new(self).tap { |klass|
```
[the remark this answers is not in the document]
timriley: Wow, TIL! This is useful!

lib/dry/system/provider/source_dsl.rb
```diff
@@ -0,0 +1,94 @@
...
+        end
+
+        attr_reader :source_class
+
+        def initialize(source_class)
+          @source_class = source_class
+        end
+
+        def setting(*args, **kwargs, &block)
+          source_class.setting(*args, **kwargs, &block)
```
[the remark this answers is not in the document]
timriley: I'd love to do this. I'm going to make a GH issue for this and will PR it separately: https://github.com/dry-rb/dry-system/issues/203

lib/dry/system/provider_sources/settings/config.rb
timriley: Good pickup, thanks :)

lib/dry/system/provider_sources/settings/config.rb
timriley: Ah yes, thanks! It was in that outer scope still as a hangover from the previous implementation, and I didn't notice the chance to move it in.

I think I might leave the loop structure as it is currently, because populating the errors hash is somewhat of a secondary effect (the primary effect is to assign the settings), and if we changed it to `errors = ...each_with_object({}) do |setting_name, errors|` it might look like the main purpose of the loop is to populate the errors only.

spec/fixtures/test/system/providers/client.rb
timriley: 👍🏼

timriley:
Thanks for the thorough feedback, @flash-gordon! ❤️

PR #209, opened 2022-01-14
Partial container exports and imports

### LINKED ISSUE #172: Add finer-grained controls around exporting/importing containers
Right now the only way to import a container is to import _everything,_ all of its registered components.

This is often undesirable: sometimes you import a container only because you want just one or two of its components.

Many container components could also quite reasonably be considered "private" or internal, and not something that other importing containers should _ever_ see.

To this end, I'd like to introduce a couple of different facilities to make container importing more intentional:

- Export privacy controls for the container's owned components. This could be a combination of magic comments (e.g. `# export: false` or `# private: true`) as well as configuration on the container
- Explicit, cherry-picking imports from other containers. Think how JS, etc. do it: `import "some.component", from: OtherContainer`, or `import "operations.*", from: OtherContainer`

Together, this would make individual containers so much more meaningful as holistic abstractions: they can encapsulate whatever they need, and expose only what should be useful to others. A container working in this sense would be very much like just another "object", but instead of public/private methods, it has public/private registrations.

This is something we will likely need to solve for Hanami 2.0 (to prevent deeply nested "application.*" registrations appearing via imported containers), so a full or partial solution to this could perhaps be a dry-system 1.0 concern, but I'd be equally fine if we found a fix in Hanami only to start with, and pushed a dry-system-level implementation of this to a post-1.0 release.

timriley:
This PR establishes support for both partial export _and_ import of dry-system containers.

## Partial exports with `config.exports`

Firstly, it allows containers to declare their exportable components, those that will be loaded into other containers when that container is imported. This is done as a simple array of keys on the container config:

```ruby
class ExportingContainer < Dry::System::Container
  configure do |config|
    config.exports = %w[
      some.public_component_a
      more.public_component_b
    ]
  end
end
```

That container can then be imported with a slightly adjusted `import` API:

```ruby
class ImportingContainer < Dry::System::Container
  import from: ExportingContainer, as: :other
end
```

When that container is imported, only the components in `config.exports` will be included. The rest of the container remains "private." The resulting keys in `ImportingContainer` would be:

```
"other.some.public_component_a"
"other.more.public_component_b"
```

By default, containers will export all their components (with `config.exports` defaulting to `nil`, representing "no exports configured"). To make a container entirely private, you can set `config.exports = []`.

## Partial imports with `import keys:`

In addition to partial exports, we also support partial _imports_, in which you provide your own list of keys when importing:

```ruby
import keys: ["some.public_component_a"], from: ExportingContainer, as: :other
```

This would import _only_ the specified key, provided it was also included in `config.exports` for the exporting container.

## Imported components are automatically protected from subsequent export

When components are imported, they are automatically protected from being subsequently exported and then imported again by another container. Imported components are considered "private" by default because they did not originate in container that imported them.

This protection applies only when no `config.exports` has been configured, i.e. if the container is exporting all components.

You can still explicitly choose to export imported components by adding them to a `config.exports` array.

A key driver for this behavior is to make it so that in applications where containers import other containers, we do not needlessly proliferate registrations and end up with containers holding multiple redundant registrations of the same component.

This situation will be present in every Hanami 2 app, for example, because each slice will automatically import a small core container under the "application" namespace. So in a slice for `Search`, you'd have access to core application components under keys like `"application.logger"`. In the current version of dry-system, if you them imported the `Search` slice into `Admin`, in the admin slice you'd end up with _both_ `"application.logger"` as well as `"search.application.logger`", both of which would resolve the same logger object.

You can of course _ignore_ that redundant registration and get on with your life, but it (a) deeply offends my own personal sensibilities, and (b) IMO gives off the impression that things aren't tied together quite right. With this change to dry-system, this will no longer be a problem! `"search.application.logger"` will not be registered when importing `Search`, because its own `"application.logger"` itself was imported, and therefore protected by default from subsequent export.

## Container loading is more efficient, and simpatico with lazy loading

Previously, any time you imported a container, that container would be finalised immediately and all its components imported.

Now, we finalize that exporting container _only_ when you're importing that container in its entirety, since this is the only way to ensure all components are loaded.

In all other cases, we leave the exporting container in lazy loading mode and import individual components as required, which fits much better behaviourally when the importing container is itself lazy loading. Otherwise we may end up finalizing a really big container just to grab one component.

Resolves #172.


lib/dry/system/container.rb
```diff
@@ -169,6 +168,9 @@ def configured?
           @__configured__.equal?(true)
         end
 
+        # @api private
+        alias_method :import_direct_namespace, :import
```
timriley, line 172: Our previous `.import` implementation still allowed  the original `Dry::Container.import` to be run by checking whether its single positional argument was a `Dry::Container::Namespace`. That's not possible with the new `.import` params signature, so I've preserved the original dry-container provided `.import` method via this alias.

TBH, though, I'm not sure whether it's worth keeping around at all. Thoughts?

timriley:
The Rubocop checks are failing for non-code-related reasons, e.g. 

```
Error: unrecognized cop Style/NumberedParametersLimit found in .rubocop.yml
Did you mean `Style/BarePercentLiterals`?
```

So let's feel free to ignore them here :)

lib/dry/system/container.rb
solnic: @timriley I don't think it is. I'd say we should reduce the codebase prior 1.0.0 final, otherwise such leftovers will haunt us later on.

solnic:
> The Rubocop checks are failing for non-code-related reasons, e.g.

@timriley BTW I'll be tweaking rubocop setup so that it doesn't cause failures and only issues warnings. I'll also make sure it doesn't check unrelated code changes (but that part is really tricky, so it may take a while).

lib/dry/system/importer.rb
```diff
@@ -38,15 +53,74 @@ def [](name)
...
+      def build_merge_container(other, keys)
+        keys.each_with_object(Dry::Container.new) { |key, ic|
+          next unless other.key?(key)
+
+          # Access the other container's items directly so that we can preserve all their
+          # options when we merge them with the target container (e.g. if a component in
+          # the provider container was registered with a block, we want block registration
+          # behavior to be exhibited when later resolving that component from the target
+          # container).
+          item = other._container[key]
```
solnic, line 110: Could you add a TODO here that it probably indicates a missing API in dry-container?

solnic (approved):

lib/dry/system/importer.rb
timriley: Will do. We actually do it in a couple of places in dry-system now, so it does feel like a part of dry-container worth expanding upon (speaking of which, we probably need to bring _that_ to 1.0 soon too).

lib/dry/system/container.rb
timriley: @solnic Cool, I might get rid of it even from this PR then. Thanks!

PR #210, opened 2022-01-16
Tidy the dependency graph plugin

timriley:
This makes some changes to make the dependency graph plugin less intrusive on the overall container behaviour. It:

- Nests its setting underneath a top-level `dependency_graph` setting
- Removes the ambiguously named and potentially buggy (due to its ivars that may end up being set after the cotnainer is frozen) `Container.strategies` method and instead provides its custom auto-injector strategies via an overload of the container's public `.injector` method options.

### DIFF
--- a/lib/dry/system/plugins/dependency_graph.rb
+++ b/lib/dry/system/plugins/dependency_graph.rb
@@ -1,7 +1,6 @@
 # frozen_string_literal: true
 
-require "dry/system/constants"
-require "dry/system/plugins/dependency_graph/strategies"
+require_relative "dependency_graph/strategies"
 
 module Dry
   module System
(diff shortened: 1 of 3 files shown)


lib/dry/system/plugins/dependency_graph.rb
```diff
@@ -12,34 +11,43 @@ module DependencyGraph
...
 
         # @api private
         def self.dependencies
-          {'dry-events': "dry/events/publisher"}
+          {"dry-events" => "dry/events/publisher"}
+        end
+
+        # @api private
+        def injector(**options)
+          super(**options.merge(strategies: DependencyGraph::Strategies))
```
flash-gordon, line 35: `super(**options, strategies: DependencyGraph::Strategies)`

timriley: Oh yes, will adjust directly in master :)

PR #215, opened 2022-01-24
Restore support for configurable instance proc

### LINKED ISSUE #180: Consider bringing back a simpler `instance` callable config for component_dirs
To allow users to do what used to be possible like this:

```ruby
auto_register!('services/wallets') do |config|
  config.instance do |component|
    if component.identifier.match?(/workers\./)
      component.loader.constant  # register job classes
    else
      component.instance # register rest components as instances
    end
  end
end
```

Which right now requires subclassing Loader:

```ruby
class AppLoader < Dry::System::Loader
  def self.call(component, *args)
    if component.identifier.match?(/workers\./)
      require!(component)
      constant(component)
    else
      super
    end
  end
end

# in Dry::System::Container config
config.component_dirs.loader = AppLoader
```

Which is a smidge less nice.

We already support "callable" options for `auto_register` and `memoize`, which are passed the `component` object, so a similar `instance` option _might_ be possible, but we'd need to make sure it either suitably _replaces_ the loader (and feels just as helpful) or gracefully augments it.

timriley:
Adds a configurable `instance` proc on all component dirs, which was previously available only on the configuration object yielded to block given to `Container.auto_register!`, removed in 0.19.0 to make way for consistently behaving and configurable component dirs.

Now users have two ways to customise component instantiation: the `instance` proc for simple cases, or a full replacement `loader` for more advanced cases.

Fixes #180.

### DIFF
--- a/lib/dry/system/component.rb
+++ b/lib/dry/system/component.rb
@@ -60,7 +60,7 @@ module Dry
       # @return [Object] component's class instance
       # @api public
       def instance(*args)
-        loader.call(self, *args)
+        options[:instance]&.call(self, *args) || loader.call(self, *args)
       end
       ruby2_keywords(:instance) if respond_to?(:ruby2_keywords, true)
 
--- a/lib/dry/system/component_dir.rb
+++ b/lib/dry/system/component_dir.rb
@@ -148,6 +148,7 @@ module Dry
         {
           auto_register: auto_register,
           loader: loader,
+          instance: instance,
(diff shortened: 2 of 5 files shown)


timriley:
@solnic @flash-gordon Are you both OK to see this come back?

flash-gordon (approved):
Yes? :)

solnic (approved):
🙇🏻

Issue #147, opened 2020-05-31
Improve clarity of the error thrown by Dry::System::Loader#constant when a constant is not found


serodriguez68:
Hello!  

I want to suggest a minor but meaningful improvement to the clarity of the error that gets raised when `Dry::System::Loader` cannot load a constant.

**Disclaimer:**  I am a dry newbie so apologies in advance if what I'm proposing doesn't make sense in the broader scheme of `dry-system`.

```
As a developer
When I create a file that will be loaded through Dry::System
And the file I created has some typo in the name of the constant (or the path)
I expect Dry::System to point me directly to the file that cannot be loaded
```

## Examples

Let's say I create a file that is meant to be injected somewhere else and I have configured my app to auto-register files living on `lib`.  However, when I create the file, I make a typo on the constant name (or on the name of any `directory` on the path to the file).

```ruby
# lib/admin/articles/commands/create_article.rb
module Admin
  module Articles
    module Command # <= I missed an "s"
      class CreateArticle
         # ...
      end
    end
  end
end
```

When I load my application, I get a `NameError` that comes from `dry-inflector`. 

```
3: from .../gems/dry-system-0.17.0/lib/dry/system/loader.rb:52:in `call'
2: from .../gems/dry-system-0.17.0/lib/dry/system/loader.rb:67:in `constant'
1: from .../gems/dry-inflector-0.2.0/lib/dry/inflector.rb:90:in `constantize'
.../gems/dry-inflector-0.2.0/lib/dry/inflector.rb:90:in `const_get': uninitialized constant Admin::Articles::Commands (NameError)
```

However, the `NameError` doesn't directly point me to the source of the problem. Instead, I get pointed to `Admin::Articles::Commands`; but I may have many files under that namespace.

### Proposed Solution

Doing something like this in `Dry::System::Loader#constant`, can make the error reporting better:

```ruby
def constant
  stringified_const_name = inflector.camelize(path)
  inflector.constantize(stringified_const_name)
rescue NameError
  raise NameError, "uninitialized constant #{stringified_const_name}"
end
```

```
gems/dry-system-0.17.0/lib/dry/system/loader.rb:69:in `rescue in constant': uninitialized constant Admin::Articles::Commands::CreateArticle (NameError)
```

## Extra Notes
- I don't know if this should be implemented here or in `dry-inflector#constantize`, but I think it makes more sense in the context of `dry-system`.

cc @timriley

timriley:
[...]

PR #217, opened 2022-01-25
Raise ComponentNotLoadableError with helpful dry-system messages

cllns:
Addresses #147, with "Did you mean?" support

### DIFF
--- a/lib/dry/system/errors.rb
+++ b/lib/dry/system/errors.rb
@@ -92,5 +92,44 @@ module Dry
         super("dry-system plugin #{plugin.inspect} failed to load its dependencies: #{details}")
       end
     end
+
+    # Exception raised when auto-registerable component is not loadable
+    #
+    # @api public
+    ComponentNotLoadableError = Class.new(NameError) do
+      # @api private
+      def initialize(component, error,
+                     corrections: DidYouMean::ClassNameChecker.new(error).corrections)
+        full_class_name = [error.receiver, error.name].join("::")
+
+        message = [
+          "Component '#{component.key}' is not loadable.",
+          "Looking for #{full_class_name}."
+        ]
+
+        if corrections.any?
+          case_correction = corrections.find { |correction| correction.casecmp?(full_class_name) }
+          if case_correction
+            acronyms_needed = case_correction.split("::").difference(full_class_name.split("::"))
+            stringified_acronyms_needed = acronyms_needed.map { |acronym|
+              "'#{acronym}'"
+            } .join(", ")
+            message <<
+              <<~ERROR_MESSAGE
+
+                You likely need to add:
+
+                    acronym(#{stringified_acronyms_needed})
+
+                to your container's inflector, since we found a #{case_correction} class.
+              ERROR_MESSAGE
+          else
+            message << DidYouMean.formatter.message_for(corrections)
+          end
+        end
+
+        super message.join("\n")
+      end
+    end
   end
 end
--- a/lib/dry/system/loader.rb
+++ b/lib/dry/system/loader.rb
@@ -1,5 +1,7 @@
 # frozen_string_literal: true
 
+require "dry/system/errors"
+
 module Dry
   module System
     # Default component loader implementation
@@ -61,7 +63,10 @@ module Dry
         # @api public
         def constant(component)
           inflector = component.inflector
-          inflector.constantize(inflector.camelize(component.const_path))
+          const_name = inflector.camelize(component.const_path)
+          inflector.constantize(const_name)
+        rescue NameError => e
+          raise ComponentNotLoadableError.new(component, e)
         end
 
         private
--- a/spec/unit/auto_registrar_spec.rb
+++ b/spec/unit/auto_registrar_spec.rb
@@ -1,6 +1,7 @@
 # frozen_string_literal: true
 
 require "dry/system/auto_registrar"
+require "dry/system/errors"
 
 RSpec.describe Dry::System::AutoRegistrar, "#finalize!" do
   let(:auto_registrar) { described_class.new(container) }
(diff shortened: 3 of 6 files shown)


solnic (approved):

solnic:
@cllns could you address Rubocop comments + squash commits into one?

flash-gordon:
```
Naming/RescuedExceptionsVariableName: Use `e` instead of `error`.
```
loool

timriley (changes requested):
This is a good place to be catching errors and raising more tailored ones, thanks!

Left a small naming suggestion, and a question about the effectiveness of the DidYouMean integration in its current state. Keen to hear what you think 😄

lib/dry/system/errors.rb
```diff
@@ -92,5 +92,36 @@ def initialize(plugin, message, gem = nil)
...
+      def initialize(component, error, corrections: DidYouMean::ClassNameChecker.new(error).corrections)
+        full_class_name = [error.receiver, error.name].join("::")
+
+        message = [
+          "Component '#{component.key}' with auto_register: true is not loadable.",
+          "Looking for #{full_class_name}."
+        ]
+
+        if corrections.any?
+          case_correction = corrections.find { |correction| correction.casecmp?(full_class_name) }
```
cllns, line 110: TIL `String#casecmp?` :)

lib/dry/system/errors.rb
```diff
@@ -92,5 +92,36 @@ def initialize(plugin, message, gem = nil)
...
+          if case_correction
+            acronyms_needed = case_correction.split("::").difference(full_class_name.split("::"))
```
cllns, line 112: `Array#difference` isn't very popular method, but... it's what we need! 

By doing it this direction (instead of the opposite), we get the correct existing (acronym-ized) part of the namespace. If we do it the other direction, we'd need to `.upcase` the result.

lib/dry/system/loader.rb
```diff
@@ -61,7 +63,10 @@ def call(component, *args)
         # @api public
         def constant(component)
           inflector = component.inflector
-          inflector.constantize(inflector.camelize(component.const_path))
+          const_name = inflector.camelize(component.const_path)
```
cllns, line 66: Oops, this is leftover from some other changes I had. Happy to get rid of it, but I find it does make the code a little more readable

spec/unit/auto_registrar_spec.rb
```diff
@@ -22,6 +23,20 @@
     expect(container["foo"]).to be_an_instance_of(Test::Foo)
     expect(container["bar"]).to be_an_instance_of(Test::Bar)
     expect(container["bar.baz"]).to be_an_instance_of(Test::Bar::Baz)
+
+    expect { container["bar.abc"] }.to raise_error(
+      Dry::System::ComponentNotLoadableError
+    ) do |error|
+      expect(error.message).to eq(
```
cllns, line 30: Should I get rid of this message expectation? We already test the message in the new `spec/unit/dry/system/errors_spec.rb` file

spec/unit/loader_spec.rb
```diff
@@ -117,4 +117,44 @@ module Test
...
+    describe "unsuccessful constant loading" do
+      before do
+        Test::APIBoo = Class.new
+      end
+
+      it "raises custom error" do
+        expect { loader.constant(component) }.to raise_error(
+          Dry::System::ComponentNotLoadableError
+        ) do |error|
+          expect(error.message).to eq(
```
cllns, line 148: Should I get rid of this message expectation? We already test the message in the new `spec/unit/dry/system/errors_spec.rb` file

spec/unit/auto_registrar_spec.rb
```diff
@@ -22,6 +23,17 @@
...
+
+    expect { container["bar.abc"] }.to raise_error(
+      Dry::System::AutoRegisterableComponentNotLoadable
+    ) do |error|
+      expect(error.message).to start_with(
+        "Component 'bar.abc' with auto_register: true is not loadable. "\
+        "Looking for Test::Bar::Abc. You may need customize your inflector.\n"\
+      ).and end_with(
+        "Did you mean?  Test::Bar::ABC"
+      )
```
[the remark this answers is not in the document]
cllns: Yes, thanks for this!

Defining our own Checker doesn't really do much, since the message formatting is elsewhere. Instead, we can just use the existing checker's `corrections` array to get what we're looking for, then constructing our own string based on that.

lib/dry/system/loader.rb
timriley: Yeah, this is fine by me :)

lib/dry/system/errors.rb
```diff
@@ -92,5 +92,36 @@ def initialize(plugin, message, gem = nil)
...
+    # Exception raised when auto-registerable component is not loadable
+    #
+    # @api public
+    ComponentNotLoadableError = Class.new(NameError) do
+      # @api private
+      def initialize(component, error, corrections: DidYouMean::ClassNameChecker.new(error).corrections)
+        full_class_name = [error.receiver, error.name].join("::")
+
+        message = [
+          "Component '#{component.key}' with auto_register: true is not loadable.",
```
timriley, line 105:
```suggestion
          "Component '#{component.key}' is not loadable.",
```

Just noticed this (sorry!), and consistent with the thoughts I left before, I reckon we could even do away with the mention of "auto_register: true" inside the error message, since I think this will be implicitly clear to our users, and in most cases, they won't be explicitly mentioning `auto_register: true` on the components, so this detail might be more confusing than not.

timriley (commented):
Really love the changes you made to this, @cllns, thanks for continuing with it! Just left one last note about a tweak in the error message, then we can get this thing in :)

lib/dry/system/errors.rb
cllns: Ah yes, of course!

timriley:
Wonderful! Thank you @cllns 🙏🏼

PR #218, opened 2022-01-30
Conditional providers

### LINKED ISSUE #94: Support conditional booting of providers

timriley:
Allow providers to be conditionally registered based on a new `if:` argument for `Container.register_provider`.

The value for `if:` is expected to be a simple truthy or falsey value only. It defaults to `true` when not provided.

This feature is valuable because it allows certain providers, which may only be needed in certain runtime contexts (i.e. when the container is left lazy loading and those providers or dependent components are accessed directly) to be skipped when finalizing the container to run the "main" parts of the app. For example, there may be a provider required purely for a subset of background jobs, which will be loaded when running the jobs worker, but is not needed at all when the main web application is run via Puma and a finalised container.

In this PR we also move all provider-registration-related methods and logic from `Container` into `ProviderRegistrar`, which helps to make `Container` considerably shorter, while also keeping all provider registration logic together inside `ProviderRegistrar`.

In terms of implementation, I toyed for a moment (see [this commit](https://github.com/dry-rb/dry-system/pull/218/commits/196fe2985c29523cec7d487ddc866cd24746687e)) with allowing a proc to be passed here, but given the implementation had the proc immediately evaluated (and passed the container object, which the user already has access to given it's the receiver for `register_provider`), I feel it wasn't adding any value. For this feature I think I'd rather keep it simple to start with, and then consider evolution of the behavior based on real experience and user feedback.

Resolves #94, resolves #95

### DIFF
--- a/lib/dry/system/container.rb
+++ b/lib/dry/system/container.rb
@@ -290,7 +290,7 @@ module Dry
         # @return [self]
         #
         # @api public
-        def register_provider(name, namespace: nil, from: nil, source: nil, &block)
+        def register_provider(name, namespace: nil, from: nil, source: nil, if: true, &block)
           raise ProviderAlreadyRegisteredError, name if providers.key?(name)
 
           if from && source.is_a?(Class)
@@ -300,6 +300,9 @@ module Dry
           if block && source.is_a?(Class)
             raise ArgumentError, "You must supply only a `source:` option or a block, not both"
           end
+
+          provider_if = binding.local_variable_get(:if)
+          return self unless provider_if
 
           provider =
             if from
@@ -315,6 +318,8 @@ module Dry
             end
 
           providers.register_provider provider
+
+          self
         end
         # rubocop:enable Metrics/PerceivedComplexity
 
--- a/lib/dry/system/container.rb
+++ b/lib/dry/system/container.rb
@@ -8,7 +8,6 @@ require "dry-container"
 require "dry/core/deprecations"
 require "dry/inflector"
 
-require "dry/system"
 require "dry/system/auto_registrar"
 require "dry/system/component"
 require "dry/system/constants"
@@ -218,8 +217,6 @@ module Dry
           self
         end
 
-        # rubocop:disable Metrics/PerceivedComplexity
-
         # Registers a provider and its lifecycle hooks
         #
         # By convention, you should place a file for each provider in one of the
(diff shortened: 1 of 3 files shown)


lib/dry/system/provider_registrar.rb
```diff
@@ -36,12 +37,43 @@ def freeze
...
+
+        if from && source.is_a?(Class)
+          raise ArgumentError, "You must supply a block when using a provider source"
+        end
+
+        if block && source.is_a?(Class)
+          raise ArgumentError, "You must supply only a `source:` option or a block, not both"
+        end
+
+        return self unless binding.local_variable_get(:if)
```
timriley, line 55: This line here is literally the only new behaviour we're adding.

solnic (approved):

lib/dry/system/container.rb
```diff
@@ -218,106 +217,101 @@ def import(keys: nil, from: nil, as: nil, **deprecated_import_hash) # rubocop:di
...
-        #   #
-        #   # Provider registration which uses another provider
-        #   MyApp.register_provider(:db) do |container|
-        #     use :logger
-        #
-        #     start do
-        #       require "db"
-        #       DB.configure(ENV['DB_URL'], logger: logger)
-        #       container.register("db", DB.new)
+        #     # system/boot/db.rb
```
flash-gordon, line 279: should it be "system/providers/db.rb"?

PR #219, opened 2022-01-31
Add Identifier#end_with? and #include?

### LINKED ISSUE #71: Share an example within a gem?
Hey guys,

I'm having a hard time configuring dry-system within a new gem context. The 2 examples provided in this repo are pretty close but they lack the fact that when creating a new gem, the lib folder look something like this:

```
lib
├── foo
│   └── version.rb
└── foo.rb
```

Where we can then namespace services objects using dry-transactions within a `Transactions` module:

```
lib
├── foo
│   ├── transactions
│   │   └── do_something.rb
│   └── version.rb
└── foo.rb
```

But i'm confused on hot to setup the `configure` hook of dry-system. This is what I have so far:

```ruby
# system/container.rb
module Foo
    configure do |config|
      config.root = Pathname(File.dirname(__FILE__)).join('../../')
      config.auto_register = %w(lib/foo/transactions)
    end

    load_paths!('lib/foo')
  end
end
```

but when testing in console with:

```ruby
Foo::Container['transactions.do_something']
NameError: uninitialized constant Transactions::DoSomething
```

It's missing the gem namespace  `Foo`.
Any idea what i'm missing?

Would be nice to include in this repo a full example using a gem as the file structure, what do you think?

### LINKED ISSUE #171: Add more useful namespace-aware predicate methods to `Dry::System::Identifier`
Right now we have just `#start_with?`, but it would be good to have at least `#include?` too, and perhaps some others may also make sense.

timriley:
This expands upon our range of helpful `Identifier` predicates and resolves #171.

### DIFF
--- a/lib/dry/system/identifier.rb
+++ b/lib/dry/system/identifier.rb
@@ -67,6 +67,28 @@ module Dry
           key.eql?(leading_namespaces)
       end
 
+      # Returns true if the given trailing segments string is the end part of the
+      # identifier's key.
+      #
+      # Also returns true if nil or an empty string is given.
+      #
+      # @example
+      #   identifier.key # => "articles.operations.create"
+      #
+      #   identifier.end_with?("create") # => true
+      #   identifier.end_with?("operations.create") # => true
+      #   identifier.end_with?("ate") # => true
+      #
+      # @param trailing_segments [String] the one or more trailing key segments to check
+      # @return [Boolean]
+      # @api public
+      def end_with?(trailing_segments)
+        trailing_segments.nil? ||
+          trailing_segments.empty? ||
+          key.end_with?("#{KEY_SEPARATOR}#{trailing_segments}") ||
+          key.eql?(trailing_segments)
+      end
+
       # Returns the key with its segments separated by the given separator
       #
       # @example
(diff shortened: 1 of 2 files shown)


timriley:
@flash-gordon In #171 you suggested `#match?` as a possible key segment-aware predicate that we could add to `Identifier`.

After adding `#end_with?` and `#include?` here (both of which have fairly obvious behaviours around making them key segment-aware), I'm not really sure what would be the right thing to do for `#match?`.

Given a key of `"foo.bar.baz"`, would you want e.g. `#match?(pattern?` to return true if any individual segment matched? So `match?("ba") # => true` and `match?(/.oo/) # => true` but `match?(bar.*z) # => false`?

flash-gordon:
@timriley I took a look into my apps, I don't have a place where I would need a `match?` against a full key. Except for one place where `match?(/_plugins\./)` matches both `rom_plugins` and `roda_plugins`. So, I think `.include?` is enough. As an option, we can make `start_with?`/`end_with?` work with paths: `.start_with?('top.middle')` or `.start_with?('top', 'middle')`.

flash-gordon:
Ah, now I took a look at the sources, it already works this way, nevermind :)

flash-gordon (approved):

timriley:
Thanks for checking this, @flash-gordon! And I was happy that the existing code already satisfied your wishes 😄

PR #221, opened 2022-01-31
Add `:injector_mixin` plugin

timriley:
The `:injector_mixin` plugin makes it an easy one-liner to define a constant for the given container's auto-injector mixin.

This way you never have to worry about doing `Deps = MyContainer.injector` anywhere.

It looks like this:

```ruby
module MyApp
    use :injector_mixin

    configure do |config|
      # ...
    end
  end
end
```

By this point, there will now be a `MyApp::Deps` defined, which you can mix into your component classes as usual, e.g. `include Deps["some_dependency"]`.

You can also customise the name:

- `use :injector_mixin, name: "Inject"` will create `MyApp::Inject`
- `use :injector_mixin, name: "Nested::Inject"` will create `MyApp::Nested::Inject`
- `use :injector_mixin, name: "::Inject` will create `::Inject`

The mixin constant is defined in the container's after-configure hook, since we rely on the inflector for camelizing and constantizing, and we want to give the user an opportunity to provide a custom inflector.


timriley:
My big question here: what do you think about the _name?_ I think `injector_mixin` is probably the most self-descriptive, but I'm not totally wedded to it, so I'm open to other suggestions at this point.

lib/dry/system/plugins/injector_mixin.rb
```diff
@@ -0,0 +1,65 @@
...
+    module Plugins
+      # @api private
+      class InjectorMixin < Module
+        MODULE_SEPARATOR = "::"
+
+        attr_reader :name
+
+        def initialize(name: "Deps")
+          @name = name
+        end
```
timriley, line 14: Rubocop is saying:

```
Lint/MissingSuper: Call `super` to initialize state of the parent class.
```

Do we really think we need to do this here?

francois: I have a feeling we should always delegate to `super`. Who knows where this class can end up being used later? If it doesn't properly delegate, this would break expectations.

timriley: My question is more about whether it's really necessary when inheriting from `Module`. We already don't do it across the other instances of `Module` inheritance in this codebase, and it's not something that's appeared necessary to me when using this technique elsewhere.

flash-gordon: @timriley I doubt it plays a role in this particular case. It's not a problem to have `super` calls to quell rubocop, though

flash-gordon:
can I use different strategies with this? I usually have something like
```ruby
    Import = injector(dynamic: env.eql?('test')) # default strategy is effects provided by Dry::Effects::System::Container
    KwargsImport = injector(effects: false)      # here opt out back to kwargs
```

timriley:
@flash-gordon hmm, no! I've not seen this kind of advanced use case before. Would you have ideas about if/how we could support it?

flash-gordon:
API-wise I'd expect `use :injector_mixin, dynamic: ENV['APP_ENV'].eql?('test')` where all extra options are transparently passed to the injector builder (i.e. `container.injector`).

lib/dry/system/plugins/injector_mixin.rb
solnic: I think it's beneficial to have a habit of calling super, even if it's a noop. It may happen that you really need to call super, but you don't have this habit and you end up debugging why things don't work just realize you forgot to call super. Been there done that 🙂

PR #231, opened 2022-02-16
[Proof of concept] Component isolation

cllns:
dry-system provides a nice container for creating new instances of classes, but it doesn't enforce its use. Developers are free to create an object by calling its constant constructor (e.g. `Foo.new`)

_...until now..._ 😱 

With this proposed 'isolation' feature, a component's constant would be removed right after it's loaded, so the users would **not** be able to call `Foo.new`. Or call `Foo` at all. That is, they would _have_ to use the container to create new instances.

This may be a very bad idea... it may be going _too far_. It may have negative unintended consequences. I could see it being useful in some circumstances though. The idea is to remove more global state, by having fewer constants.

This is clearly just a proof of concept. It could be a post-1.0 feature, too. If we wanted to implement this for real, some things (among more, I'm sure) we'd have to consider:

- How would this be configured? Per component, per component_dir, per container? 
- Should this be a custom `Loader`? i.e. provided as a subclass of Loader, either in this gem or as an external one
- Is this the right place to do this?
- Is isolate/isolation the right name?
- Do we want to go further, and load the class definitions once and store them internally somehow, and remove the constants then, once?

### DIFF
--- a/lib/dry/system/loader.rb
+++ b/lib/dry/system/loader.rb
@@ -41,16 +41,25 @@ module Dry
         # @return [Object]
         #
         # @api public
-        def call(component, *args)
+        def call(component, *args, isolate: false)
           require!(component)
 
           constant = self.constant(component)
 
-          if singleton?(constant)
+          instance = if singleton?(constant)
             constant.instance(*args)
           else
             constant.new(*args)
           end
+
+          if isolate
+            constant_name_parts = constant.to_s.split("::")
+            namespace = constant_name_parts.slice(0..-2).join("::")
+            namespace_const = component.inflector.constantize(namespace)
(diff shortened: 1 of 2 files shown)


cllns:
This can also be implemented as a custom Rubocop rule:

```
module RuboCop
  module Cop
    module Experimental
      class ProhibitClassNew < Base
        MSG = 'Use your dependency injection framework instead of manually creating objects from class constants.'

        RESTRICT_ON_SEND = %i[new].freeze

        def_node_matcher :class_new?, <<~PATTERN
          (send const :new ...)
        PATTERN

        def on_send(node)
          add_offense(node) if class_new?(node)
        end
        alias on_csend on_send
      end
    end
  end
end
```

PR #234, opened 2022-05-05
Raise error when import is called on finalized container

### LINKED ISSUE #228: Importer should raise errors if new imports are registered after finalize
This would also mean we don't need to worry about handling this error from the Hanami side of dry-system usage.

tak1n:
Let me know whether the error should have another name, I thought the name `ContainerAlreadyFinalizedError` is adequate as it can be reused in the future for any operation on a finalized container which should raise.

I've also found the yard doc on `Dry::System::Container.import` is outdated, let me know if I should update that. Found the description in PR https://github.com/dry-rb/dry-system/pull/209 quite helpful and I could start to transfer some bits and pieces of it into the doc.

Closes #228

### DIFF
--- a/lib/dry/system/container.rb
+++ b/lib/dry/system/container.rb
@@ -171,26 +171,31 @@ module Dry
         #
         # @example
         #   # system/container.rb
+        #   require "dry/system/container"
+        #   require "logger"
+        #
         #   class Core < Dry::System::Container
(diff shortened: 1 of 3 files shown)


timriley:
Thanks for taking care of this, @tak1n! I've resolved a merge conflict in #254 and brought this in :)

PR #235, opened 2022-06-13
Fix incorrect type in ManifestRegistrar#finalize!

alassek:
`#finalize!` is sending the basename directly to `#call`, which expects a `Dry::System::Identifier`.

This means that if you configure a `registrations_dir`, putting any file in that directory will lead to an exception during finalization.

This appears to have been introduced during #181


solnic (commented):
Thanks for addressing this. Do you think you could add a corresponding test too? 🙏🏻

alassek:
I added a test to the integration spec that triggers the bug. Also fixed the Zeitwerk failures due to a private interface change.

solnic:
Thank you ❤️

PR #236, opened 2022-06-22
Import root components via `nil` import namespace

timriley:
With this change, we can import components from one container to the root of another by providing a namespace of `nil` to the `import` directive:

```ruby
  # ...

  import from: SomeOtherContainer, as: nil
end
```

Generally, having your imports namespaced is a good practice, because it serves as a clear hint as to their origin. However, in some cases, it may be useful not to have a namespace. One such case is in our plans for Hanami 2.0, where we want a small selection of application-wide components (such as `"logger"` and `"inflector"`) to be available to every container with the same keys, so there can a consistency in the way they are addressed.

### DIFF
--- a/lib/dry/system/container.rb
+++ b/lib/dry/system/container.rb
@@ -193,7 +193,7 @@ module Dry
         # @param other [Hash, Dry::Container::Namespace]
         #
         # @api public
-        def import(keys: nil, from: nil, as: nil, **deprecated_import_hash)
+        def import(keys: nil, from: Undefined, as: Undefined, **deprecated_import_hash)
           if deprecated_import_hash.any?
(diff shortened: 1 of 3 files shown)


solnic (approved):
Looking good!

flash-gordon:
Shouldn't be this the default for `as:`? On the one hand, passing `nil` explicitly is good because it's visible, on the other hand, I would expect something more verbose (`as: :root`, it would require treating symbols and strings separately). It's just a thought, I'm good with `as: nil` too.

solnic:
When it comes to ergonomics, I thought about it too and what would make sense to me, would be to make `as: nil` the default.

PR #243, opened 2022-07-09
Fix compat with dry-container from main

solnic:
This makes things work again with dry-container from main:

- Use the new `Dry::Container::KeyError` (`Error` is deprecated)
- Require `dry/configurable` before `dry/container` so that `Configurable` is used as the config backend


jodosha (approved):
@solnic Thanks for taking care of this. 👍

PR #244, opened 2022-07-27
Use autoloading for dry-core and dry-configurable

flash-gordon:
I think I managed to make it work without warnings. We should merge this PR once we're ready to cut a release.


flash-gordon:
Nah, specs are failing with certain seeds, I'll investigate.

flash-gordon:
Fixed

solnic (approved):

solnic:
I'd merge this in already and then follow-up with another one that zeitwerkifies dry-system's own requires. WDYT?

flash-gordon:
@solnic pls go ahead

PR #246, opened 2022-09-05
Validate settings even if loader does not set value

### LINKED ISSUE #245: Missing settings are not validated
## Describe the bug
When the Setting::Loader is unable to determine the value of a particular setting it silently fails (does not raise an `InvalidSettingsError`). The setting value is only evaluated at runtime when some downstream component attempts to access it, and the exception raise is a `Dry::Types::ConstraintError` which makes no reference to the name of the offending/missing setting making it very hard to debug.

## To Reproduce

Declare a settings provider that inherits from the settings provider source:

```ruby
App.register_provider(:db_config, from: :dry_system, source: :settings) do
  settings do
    setting(:db_host, constructor: Types::String.constrained(filled: true))
  end
end
```

Then don't set a `DB_HOST` var in the environment.

## Expected behavior

I get an `InvalidSettingsError` when the settings are loaded. Something like:

```
Could not load settings. The following settings were invalid: (Dry::System::ProviderSources::Settings::InvalidSettingsError)
db_host: nil violates constraints (type?(String, nil) AND filled?(nil) failed)
```

## My environment

- Affects my production application: Not yet
- Ruby version: reproduced on 2.7.5 and 3.1.2
- OS: macOS

oeoeaio:
Resolves #245 

This is probably not the right way to solve this, but it gives the intended effect.

### DIFF
--- a/lib/dry/system/provider_sources/settings/config.rb
+++ b/lib/dry/system/provider_sources/settings/config.rb
@@ -41,7 +41,11 @@ module Dry
                 value = loader[setting_name.to_s.upcase]
 
                 begin
-                  settings_obj.config.public_send(:"#{setting_name}=", value) if value
+                  if value
+                    settings_obj.config.public_send(:"#{setting_name}=", value)
+                  else
+                    settings_obj.config[setting_name]
+                  end
                 rescue => e # rubocop:disable Style/RescueStandardError
                   errors[setting_name] = e
                 end
--- a/spec/integration/deprecations/deprecated_settings_provider_source_key_spec.rb
+++ b/spec/integration/deprecations/deprecated_settings_provider_source_key_spec.rb
@@ -30,10 +30,16 @@ RSpec.describe "Deprecated :settings provider source `key`" do
     Test::Container
   }
 
-  it "defines the setting" do
+  before do
     ENV["SOME_INT_USING_DEPRECATED_KEY"] = "5"
-    expect(container[:settings].some_int_using_deprecated_key).to eq 5
+  end
+
+  after do
     ENV.delete("SOME_INT_USING_DEPRECATED_KEY")
+  end
+
+  it "defines the setting" do
+    expect(container[:settings].some_int_using_deprecated_key).to eq 5
   end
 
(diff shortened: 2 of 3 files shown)


spec/integration/deprecations/deprecated_settings_provider_source_key_spec.rb
```diff
@@ -30,12 +30,18 @@ class Container < Dry::System::Container
...
+  end
+
+  after do
     ENV.delete("SOME_INT_USING_DEPRECATED_KEY")
   end
 
+  it "defines the setting" do
+    expect(container[:settings].some_int_using_deprecated_key).to eq 5
+  end
+
```
oeoeaio, line 44: Had to tweak this because this spec 👇 was failing with a validation error.

lib/dry/system/provider_sources/settings/config.rb
```diff
@@ -41,7 +41,11 @@ def self.load(root:, env:, loader: Loader)
                 value = loader[setting_name.to_s.upcase]
 
                 begin
-                  settings_obj.config.public_send(:"#{setting_name}=", value) if value
+                  if value
+                    settings_obj.config.public_send(:"#{setting_name}=", value)
+                  else
+                    settings_obj.config[setting_name]
+                  end
```
oeoeaio, line 48: Basically, if the loaded doesn't find a value, we still evaluate the setting to make sure the constructor is happy with the setting being `nil`.

oeoeaio:
@solnic is there anything I can do to move this along?

timriley:
@oeoeaio Thank you for submitting this, and sorry for the delay here! I've actually just addressed this same issue in Hanami (https://github.com/hanami/hanami/pull/1222), which does not use the dry-system settings plugin but does a very similar thing.

I can commit to checking out your PR sometime in the next ~week and have it be released.

oeoeaio:
That would be great, thanks @timriley.

timriley:
Thanks for taking care of this, @oeoeaio! This fix is perfect :)

PR #249, opened 2022-10-05
Update for dry-configurable settings separation

timriley:
The main change here is that we stop relying on `Dry::Configurable::Config` internal implementation details by using the new `#configured?` API introduced in https://github.com/dry-rb/dry-configurable/pull/143.

While we're here, we make one other small tweak to improve dry-configurable compatibility: use dry-configurable's class-level `#finalize` instead of `Config#finalize` directly, which will mean that we'll continue to have correct behaviour if dry-configurable ever moves any additional logic to this class-level method.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -6,6 +6,8 @@ eval_gemfile "Gemfile.devtools"
 
 gemspec
 
+gem "dry-configurable", github: "dry-rb/dry-configurable"
+
 # Remove verson constraint once latter versions release their -java packages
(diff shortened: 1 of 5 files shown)


timriley:
Since this is a fairly rote PR, I intend to merge it in a day or so (i.e. next time I'm working on this) to get ready for releasing both dry-configurable and dry-system.

flash-gordon (approved):

PR #255, opened 2022-10-26
Remove all deprecated functionality and deprecation messages

timriley:
Our next release will be 1.0.0.rc1, and we do not want to carry these deprecations going forward.


solnic:
@timriley must've felt goooood 😀

timriley:
@solnic SO. GOOD. 😆

PR #256, opened 2022-10-30
Switch to Dry::Core::Container

solnic:
Follow-up after dry-rb/dry-core#77


solnic:
Notice that dry-types still pull in dry-container and things do not break, which was a good accidental exercise

flash-gordon (approved):
It won't help since we have to update dry-core dependencies everywhere anyway. I don't think it's a major problem, though. My plan is to merge this one, update all the other gems, and test (= deploy) everything

Issue #253, opened 2022-10-24
Allow custom environment variable prefixes in settings provider


jeremiahrose:
Following on from the discourse conversation [here](https://discourse.dry-rb.org/t/dry-system-settings-provider-source-custom-environment-variable-prefix/1510).

When using the settings provider source, it currently isn't possible to assign a prefix to an environment variable name without also having the prefix in the internal setting name. We have written a patch and it seems straightforward enough to allow an optional prefix to be configured in the `register_provider` block like so:

```
Application.register_provider(:settings, from: :dry_system) do
  prefix { 'SOME_PREFIX_' }

  settings do
    setting :database_url, constructor: Types::String.constrained(filled: true)
end
```

I would like to submit a PR for this soon, if maintainers feel it would be helpful!

@solnic also mentioned that:
> Having ability to have nested keys from env would be great too like STRIPE__API_KEY would be turned into settings.stripe.api_key.

I agree that would be a nice feature, so I would be happy to implement that too.

jeremiahrose:
Sorry about `bug` label, I can't seem to remove it. :(

jeremiahrose:
@alassek we went for a configuration option instead, PR up here: https://github.com/dry-rb/dry-system/pull/258

PR #258, opened 2022-11-02
Allow env variable prefix and setting namespace to be configured in settings provider

jeremiahrose:
Closes https://github.com/dry-rb/dry-system/issues/253

This PR adds 2 new configuration options to the settings provider source:

1. `prefix`: sets the prefix of the environment variables that will be read. E.g with a prefix of `APP1_` and a setting named `log_level`, the settings provider will look for an environment variable named `APP1_LOG_LEVEL`.

2. `register_as`: sets the namespace that the setting will be registered to. This enables settings objects to be organised according to their purpose e.g `Container['app.config.database']` and also allows multiple settings providers to be registered.

Update: rebased to latest `main`

### DIFF
--- a/docsite/source/settings.html.md
+++ b/docsite/source/settings.html.md
@@ -19,6 +19,10 @@ Application.register_provider(:settings, from: :dry_system) do
     require "your/types/module"
   end
 
+  configure do |config|
+    config.prefix = 'SOME_PREFIX_'
+  end
+
   settings do
     setting :database_url, constructor: Types::String.constrained(filled: true)
 
@@ -29,7 +33,7 @@ Application.register_provider(:settings, from: :dry_system) do
 end
 ```
 
-Your provider will then map `ENV` variables to a struct object giving access to your settings as their own methods, which you can use throughout your application:
+An optional prefix can be specified with the `config.prefix` setting inside a `configure` block. Your provider will then map `ENV` variables with the given prefix to a struct object giving access to your settings as their own methods, which you can use throughout your application:
 
 ```ruby
 Application[:settings].database_url # => "postgres://..."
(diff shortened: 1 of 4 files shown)


solnic:
Thank you for working on this but this has to wait. We're in the process of releasing dry-system 1.0.0 and then Hanami 2.0.0 and it's code freeze now. We'll get back to this PR after we're done with the Hanami release (so, before the end of the month).

timriley:
Thanks @jeremiahrose for this contribution, and thanks @solnic for saying what I was planning to get around to saying too :)

In terms the viability of this feature, I'd definitely be confident in it being a reasonable post-1.0.0 addition. Nothing here e.g. looks like it'd be a breaking change to any of the APIs release to 1.0.0, so it'd be a straightforward expansion/addition of our existing APIs.

However, before merging anything, I expect I'd probably want to have a think about different possible ways of satisfying the end goal here. One thing I worry about with adding built-in support for very specific features (like a variable prefix) is that if we continue that same approach for any other adjustments people want with settings loading, we may end up continuing to glom things on and end up with a hard-to-understand combination of different configs around settings loading (think the [homer car](https://simpsons.fandom.com/wiki/The_Homer), and honestly dry-system is probably already a little like that, so I'd like to be quite intentional in the way we add additional features from here).

For example, another option here could be to instead add support for a single configurable "settings store" (which is something we actually have for Hanami's settings implementation). By default, this would pulling settings straight from the env like we do now. But this would also give us a way to satisfy your needs through providing a custom "prefixed names settings store" that pulls from the env using a provided prefix. This kind of arrangement may ultimately be more flexible to serving different needs without us having to add too much complexity to this part of dry-system.

jeremiahrose:
> another option here could be to instead add support for a single configurable "settings store" (which is something we actually have for Hanami's settings implementation). By default, this would pulling settings straight from the env like we do now. But this would also give us a way to satisfy your needs through providing a custom "prefixed names settings store" that pulls from the env using a provided prefix.

I guess that could tie in with the current changes quite well, just make the "prefixed names settings store" the default one. I think being able to set environment variable prefixes is a fairly standard feature for an environment loader that most users will want.

jeremiahrose:
Congrats on the release of dry-system 1.0! Is there anything I can do to help move this card along?

jeremiahrose:
Just wondering if `dry-system` is still in code freeze and is there anything I can do to help move this card along?

PR #260, opened 2022-11-07
Remove usage of the private Zeitwerk::Registry.loaders

### LINKED ISSUE #259: Zeitwerk::Registry is private interface
Hi!

The [`Zeitwerk::Registry`](https://github.com/fxn/zeitwerk/blob/main/lib/zeitwerk/registry.rb) class is private interface. It is undocumented, and its internal interface is marked as `@private` all over the place for more stress.

I am going to rename those methods to enforce stricter rules, but wanted to understand the use case you have [here](https://github.com/dry-rb/dry-system/blob/605abd9e1f9f4c399cba0993e117e58799563bd1/spec/support/zeitwerk_helpers.rb) before doing that. Maybe it is suggesting there's a missing need.

The reason `Zeitwerk::Registry.loaders` is private has different levels. On one hand, Zeitwerk does not expose any of its internal collections directly. For example, you have `push_dir` as a method, you don't mutate a collection. The public API boundary is the one that guarantees the internal state is consistent. This is a generic public interface guideline.

However, the main point in the case of `loaders` is that gem loaders are, generally speaking, not the business of anyone else. You should not even know they exist, the way a gem loads is generally private, you just know you only need to `require` their entrypoint.

Those loaders must be protected by design, because gems must have full control over their own (unless their loaders belong to their public interface, you know).

For example, the only way to broadcast things to the registered loaders are the documented blessed ones like `Zeitwerk::Loader.eager_load`. Even then, you don't access the registry externally.

Could it be the case that you are missing a collection of loaders you control? Like, you may instantiate loaders, register them in a global collection of yours, and those are the ones you need to act upon?

fxn:
`Zeitwerk::Registry` is a private module. This is the place where Zeitwerk internally stores state that is global. For example, given this file to be required, is it managed by any loader? If yes, which one? This is all internal.

The private method `Zeitwerk::Registry.loaders` is going to disappear, and I'd like to volunteer this patch so that your test suite continues being green when it ships.

By design, Zeitwerk hides the loaders. That is, the library does not provide interface to reach that collection, and that is deliberate. If I am a gem author and I load my code with a loader, I am the only one in the entire process owning this thing. Accessing that loader from outside is not something that should be normalized by the API.

I am not saying this is going to be my design guideline forever, but changing this would be done with extra care and a clear use case that justifies it, if that ever happens. And, in any case, at most you'd get a frozen clone of the collection, never a pointer.

For frameworks or situations in which multiple pieces work together and their loaders are known to each other, you can always formalize that. Rails, for example, has two loaders, they are known, and they are public interface. In a more generic situation, if a framework has a variable collection, those loaders would register using framework API, and now that collection is yours and the design reflects it.

In the discouraged case in which you want to iterate over all loaders in memory, Ruby has builtin support in `ObjectSpace.each_object`, let's just use that. Do you want all loaders in RAM? There you have them, it is your responsibility to use that in a responsible way. This is ugly, but that is the price one pays for doing something discouraged, there's some vinegar in it.

Closes #259.

### DIFF
--- a/spec/support/zeitwerk_helpers.rb
+++ b/spec/support/zeitwerk_helpers.rb
@@ -2,17 +2,9 @@
 
 module ZeitwerkHelpers
   def teardown_zeitwerk
-    # From zeitwerk's own test/support/loader_test
-    # adjusted to work with dry-rb gem loaders
-
-    Zeitwerk::Registry.loaders.reject! do |loader|
-      test_loader = loader.dirs.any? { |dir| dir.include?("/spec/") || dir.include?(Dir.tmpdir) }
-
-      if test_loader
+    ObjectSpace.each_object(Zeitwerk::Loader) do |loader|
+      if loader.dirs.any? { |dir| dir.include?("/spec/") || dir.include?(Dir.tmpdir) }
         loader.unregister
-        true
-      else
-        false
       end
     end
   end


timriley:
Thanks @fxn, for this bitter medicine 😜

In seriousness, though, this is a great stopgap measure and I appreciate you putting it together for us. 🙏🏼

Down the track, I'd love to work with you to explore the options for grouping and later retrieving certain categories of loaders, like you mention in #259, so please feel free to keep us in the loop on your thinking there, because we can certainly try things out through  our dry-sytem/hanami test setup.

fxn:
Awesome Tim ❤️.

Absolutely, happy to work together and see if better APIs can be defined for the use cases this project or Hanami could have.

Something I'd be willing to explore is some concept of groups. In that idea, loaders by default belong to no group, but you can add the to one or more:

```ruby
Zeitwerk::Loader.new(groups: %i(dry dry-system))

Zeitwerk::Loader.each_in_group(:dry) BLOCK
```

(off the top of my head).

In a case like this, I like to have a real use case that help me see the details. If you folks believe something like this could be useful for your frameworks, more than happy to work on it :).

PR #261, opened 2022-11-15
Only use DidYouMean-integrated Error for Component loading failure

cllns:
Without this, *any* NameError (including NoMethodError) will go through the DidYouMean integration and
it will fail since that code expects only a Constant, not any name (including "undefined local variable or method")

Instead, this code will ensure we only try to do that when we get a NameError for specifically trying to load the constant we're looking for. Everything else will be re-raised and go through the normal flow (including any more DidYouMean default behavior).

Fixes hanami/hanami#1252. (Specifically see [this comment](https://github.com/hanami/hanami/issues/1252#issuecomment-1315898878))

Bug introduced in #217

Thoughts on whether we should test this? We could create a fixture file to load a component that has a NameError in it and ensure that error passes through instead of getting created as a `ComponentNotLoadableError`... but that would error out anyway. Creating a NameError in the spec just raises that before we can even get to the example.

I manually confirmed that it works with [the repo @katafrakt created to illustrate the bug](https://github.com/hanami/hanami/issues/1252#issue-1447575377).

### DIFF
--- a/lib/dry/system/loader.rb
+++ b/lib/dry/system/loader.rb
@@ -66,7 +66,12 @@ module Dry
           const_name = inflector.camelize(component.const_path)
           inflector.constantize(const_name)
         rescue NameError => e
-          raise ComponentNotLoadableError.new(component, e)
+          # Ensure it's this component's constant, not any other NameError within the component
+          if e.message =~ /#{const_name}( |\n)/
+            raise ComponentNotLoadableError.new(component, e)
+          else
+            raise e
+          end
         end
 
         private
--- a/spec/fixtures/components_with_errors/test/constant_error.rb
+++ b/spec/fixtures/components_with_errors/test/constant_error.rb
@@ -1,0 +1,5 @@
+# frozen_string_literal: true
+
+class ConstantError
+  const_get(:NotHere)
+end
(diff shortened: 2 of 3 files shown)


timriley:
@cllns This sounds and looks good!

Do you reckon there's any chance you could add a test demonstrating one of these unrelated errors _not_ getting caught up in our DidYouMean handling?

timriley:
Oh, haha, I just read your description more closely.

> We could create a fixture file to load a component that has a NameError in it and ensure that error passes through instead of getting created as a ComponentNotLoadableError

This sounds about right to me. Couldn't the test wrap the loading of that file in a before/rescure, or in an `expect { }.to raise_error` construct?

cllns:
@timriley Can you take a look at my newest commit? I haven't been able to reproduce the issue from https://github.com/hanami/hanami/issues/1252. Note that I've added `|| true` to the main conditional here so it's effectively the existing library code. I want to reproduce the existing behavior so I can be sure this fix works as desired. 

I've included comments in the `spec/integration/did_you_mean_integration_spec.rb` file. The first one, about `finalize`, isn't exactly necessary, just something that surprised me based on my (now rusty) memory of dry-system's terminology. The second one, after `system["undefined_variable"]` explains where the issue is. I guess the question is: what is Hanami doing that I'm not doing here... A lot! I guess, but what's relevant to the `Loader` not seemingly being used here?

cllns:
Thanks for finishing the spec @solnic!

PR #268, opened 2023-07-11
Use a valid constant name in the example

elia:


elia:
nvm, it was wrong only on 0.23

PR #271, opened 2024-05-17
Allow provider sources to specify provider options

timriley:
When registering a provider source, you can now provider a `provider_options:` containing a hash of default options for providers to be registered using that source.

The one provider option currently supported is `namespace:`.

Allowing a provider source to specify e.g. `provider_options: {namespace: true}` is useful if the source will register multiple components at different keys. In such a case, the provider will typically want to have a `namespace: true` provider option.

Specifying this option at the time of provider source registration then makes for a simpler and less error-prone experience when the user eventually registers a provider using that source. The user now no longer needs to remember to provide their own explicit `namespace: true`.

### DIFF
--- a/lib/dry/system/provider_registrar.rb
+++ b/lib/dry/system/provider_registrar.rb
@@ -46,7 +46,7 @@ module Dry
 
       # @see Container.register_provider
       # @api private
-      def register_provider(name, namespace: nil, from: nil, source: nil, if: true, &block)
+      def register_provider(name, from: nil, source: nil, if: true, **provider_options, &block)
         raise ProviderAlreadyRegisteredError, name if providers.key?(name)
 
         if from && source.is_a?(Class)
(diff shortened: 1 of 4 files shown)


lib/dry/system/provider_registrar.rb
```diff
@@ -46,7 +46,7 @@ def freeze
 
       # @see Container.register_provider
       # @api private
-      def register_provider(name, namespace: nil, from: nil, source: nil, if: true, &block)
+      def register_provider(name, from: nil, source: nil, if: true, **provider_options, &block)
```
timriley, line 49: We had to switch from `namespace:` as an explicit option here to having it be part of `**provider_options`, so that later on in the class (in `#build_provider_from_source`), we can use it as the first of two splats in order to let the user's provider options override the default ones from the provider source:

```ruby
        Provider.new(
          **provider_source.provider_options,
          **options,
```

Without this, we would've had to do a much more heavy-handed approach, like having `namespace: Undefined` in the params, and then building up the hash of provider args with manual handling for this (and any future) argument.

This seemed like overkill, and in a sense, it's the `Provider` class itself that is responsible for the `namespace:` argument. The job of `#register_provider` here is merely to collect that argument and pass it on, so in a sense it's better for it to be opaque here.

PR #274, opened 2024-06-15
Make Container hooks more useful for dynamic registration

alassek:
This is a (long-overdue) followup PR to [our conversation](https://discourse.hanamirb.org/t/newrelic-apm-on-hanami-2/872) related to integrating APM into a Dry-System app.

Also relates to hanami/hanami#1360

**Use Case:** I want to be able to dynamically decorate keys as they are added into the container system in order to perform APM metrics without having to touch every file.

**Use Case:** I want to audit the usage of particular namespaces in the Container using Dry-Monitor decoration, such as producing an audit log of Authorization checks.

In these use-cases, it's important that I want to react to container registration without having knowledge of precisely what keys are being registered. The decorate feature works well when keys are known in advance, but otherwise you have to enumerate the keys in the container after registration.

Enumeration is a problem, because `after_finalize` is called after the container is frozen, which is too late to decorate key values.

This PR makes two changes:
1. Establish an `after_register` hook that fires whenever a key is registered
2. Move `after_finalize` to happen prior to freezing the container.

This allows us to either decorate keys dynamically as they are registered, and to enumerate the container keys before the container is frozen in case we want to make alterations.

Due to the complexity of Container value storage between memoization and callable, I opted not to expose the key value directly; the hook code may choose to call `resolve` if it needs that.

### DIFF
--- a/docsite/source/container.html.md
+++ b/docsite/source/container.html.md
@@ -2,6 +2,8 @@
 title: Container
 layout: gem-single
 name: dry-system
+sections:
+  - hooks
 ---
 
 The main API of dry-system is the abstract container that you inherit from. It allows you to configure basic settings and exposes APIs for requiring files easily. Container is the entry point to your application, and it encapsulates application's state.
--- a/docsite/source/container/hooks.html.md
+++ b/docsite/source/container/hooks.html.md
@@ -1,0 +1,78 @@
+---
+title: Hooks
+layout: gem-single
+name: dry-system
+---
+
+There are a few lifecycle events that you can hook into if you need to ensure things happen in a particular order.
+
+Hooks are executed within the context of the container instance.
+
+### `configure` Event
+
+You can register a callback to fire after the container is configured, which happens one of three ways:
+
+1. The `configure` method is called on the container
+2. The `configured!` method is called
+3. The `finalize!` method is called when neither of the other two have been
+
+```ruby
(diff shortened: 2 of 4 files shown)


timriley (approved):
These changes look good to me! @flash-gordon — did you want to take a look at all?

> Due to the complexity of Container value storage between memoization and callable, I opted not to expose the key value directly; the hook code may choose to call resolve if it needs that.

This is a smart decision, and I agree.

timriley:
@alassek Do you think you might be able to make a "minimum viable docs" page for container hooks as part of this PR? This is a powerful feature of dry-system that currently has no coverage in docs at all.

alassek:
@timriley Good point, I'll add some.

alassek:
@timriley I've added a docsite page for container hooks. What do you think?

I was unable to think of good code examples for `configure` and `finalize` hooks; if you have any suggestions for what one might want to use them for I could flesh that out.

flash-gordon (approved):

flash-gordon:
@timriley I've read the PR and the original discussion, looks good to me

timriley:
@alassek I'm very happy with those docs! At this stage I don't think we need to block on perfect examples before getting these docs out, so I reckon we can move ahead.

I'll merge this now and have made myself a note to publish a new minor release next week.

PR #275, opened 2024-06-19
Allow Provider Sources to choose a custom superclass

alassek:
Relates to hanami/hanami#1417

The motivation for this is to allow consuming frameworks of dry-system to define their own superclass for providers, in order to add their own method apis to the class.

This is only used for user-defined providers with a block implementation.

External providers in a group do not allow this, because their superclass is defined ahead of time when they are added to the source registry. If an external provider source wants to use a different superclass, they can define a concrete class of their own instead.

The custom superclass is assumed to be a child of
Dry::System::Provider::Source.

### DIFF
--- a/lib/dry/system/provider/source.rb
+++ b/lib/dry/system/provider/source.rb
@@ -37,10 +37,20 @@ module Dry
           # @see Dry::System::Provider::SourceDSL
           #
           # @api private
-          def for(name:, group: nil, &block)
-            Class.new(self) { |klass|
+          def for(name:, group: nil, superclass: nil, &block)
+            superclass ||= self
+
+            Class.new(superclass) { |klass|
               klass.source_name name
               klass.source_group group
+
+              name_with_group = group ? "#{group}->#{name}" : name
+              klass.instance_eval <<~RUBY, __FILE__, __LINE__ + 1
+                def name
+                  "#{superclass.name}[#{name_with_group}]"
+                end
+              RUBY
+
               SourceDSL.evaluate(klass, &block) if block
             }
           end
@@ -56,14 +66,6 @@ module Dry
             if subclass.superclass == Source
               subclass.include Dry::Configurable
             end
-          end
-
-          # @api private
-          def name
-            source_str = source_name
-            source_str = "#{source_group}->#{source_str}" if source_group
-
-            "Dry::System::Provider::Source[#{source_str}]"
           end
 
           # @api private
@@ -117,7 +119,10 @@ module Dry
         #
         # @api public
         attr_reader :target_container
-        alias_method :target, :target_container
+
+        # @see #target_container
+        # @api public
+        def target = target_container
 
         # @api private
         def initialize(provider_container:, target_container:, &block)
--- a/lib/dry/system/provider_registrar.rb
+++ b/lib/dry/system/provider_registrar.rb
@@ -237,7 +237,7 @@ module Dry
         provider_source = System.provider_sources.resolve(name: source, group: group)
 
         source_options =
-          if provider_source.source == provider_source_class
+          if provider_source.source <= provider_source_class
             provider_source_options
           else
             {}
--- a/lib/dry/system/provider_registrar.rb
+++ b/lib/dry/system/provider_registrar.rb
@@ -136,6 +136,21 @@ module Dry
         }.first
       end
 
+      # Extension point for subclasses to customize their
+      # provider source superclass. Expected to be a subclass
+      # of Dry::System::Provider::Source
+      #
+      # @api public
+      # @since 1.1.0
+      def provider_source_class = Dry::System::Provider::Source
+
+      # Extension point for subclasses to customize initialization
+      # params for provider_source_class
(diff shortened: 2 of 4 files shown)


lib/dry/system/provider/source.rb
```diff
@@ -37,10 +37,20 @@ class << self
...
+            Class.new(superclass) { |klass|
               klass.source_name name
               klass.source_group group
+
+              name_with_group = group ? "#{group}->#{name}" : name
+              klass.instance_eval <<~RUBY, __FILE__, __LINE__ + 1
+                def name
+                  "#{superclass.name}[#{name_with_group}]"
+                end
+              RUBY
```
alassek, line 52: The metaprogramming around this got a bit hairy, I was unable to find an elegant way to handle this with variable superclasses, so hard-coding it this way was the least bad option.

timriley (commented):
This is really great, @alassek! Thank you so much for making it happen 😄 

I've left a couple of notes for you that should hopefully be enough for you to get this into a ready-to-merge state.

Before we merge, though, would you be up for making a matching change to hanami that uses this? I _think_ that this feature should allow us to get rid of our custom provider registrar in Hanami? That would be a nice outcome if possible — just one piece of customisation instead of two.

alassek:
@timriley 

> Before we merge, though, would you be up for making a matching change to hanami that uses this?

Certainly, I wanted to pause to get feedback on the approach before continuing, but that is the next step and shouldn't require much effort.

spec/integration/container/providers/custom_provider_superclass_spec.rb
```diff
@@ -0,0 +1,83 @@
...
+        end
+
+        def start
+          register(:logger, Logger.new($stdout))
+        end
+      end
+    end
+
+    Test::LoggerProvider
+  end
```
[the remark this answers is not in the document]
alassek: Indeed, I think this was an artifact of testing against external providers which I abandoned. ✂️

lib/dry/system/provider_registrar.rb
```diff
@@ -136,6 +136,14 @@ def provider_files
         }.first
       end
 
+      # Extension point for subclasses
+      # @api private
```
timriley, line 140:
```suggestion
      # @api public
      # @since 1.1.0
```

I think we'll want to mark these as public since we're using them from Hanami.

(Same thing applies to the other new method below).

lib/dry/system/provider_registrar.rb
```diff
@@ -215,6 +228,7 @@ def build_provider_from_source(name, source:, group:, options:, &block)
           name: name,
           target_container: target_container,
           source_class: provider_source.source,
+          source_options: provider_source_options,
```
timriley, line 231: I wonder whether we want to be more careful before passing `provider_source_options` here, because in this case, the provider registrar doesn't have the same control over the `source_class` as it does in `build_provider`, so we can't be 100% confident it will even accept any extra options.

Perhaps we want to do something like this?

```
source_options = provider_source.source == provider_source_class ? provider_source_options : {}
```

And then pass that in? This way we're only passing custom options if we know the source class matches what the registrar expects it to be.

timriley (commented):
Thanks for refreshing this PR, @alassek!

It looks like you've found a nice arrangement here, with none of the provider `build_source` proc awkwardness that I introduced in in my spike on this at https://github.com/hanami/hanami/issues/1417#issuecomment-2194542940.

So this looks good to me overall! I've just left 2 comments for you. One is superficial, but the other one (regarding `provider_source_options`) is a more important aspect of the functionality. Let me know what you think about that, and after that is sorted, we can merge this in! 😄

lib/dry/system/provider/source.rb
flash-gordon, line 52: Ruby 3.3 has `Module#set_temporary_name` https://rubyreferences.github.io/rubychanges/3.3.html#moduleset_temporary_name which would work nicely here I think

timriley:
@alassek FYI, I just pushed a commit here, which made the Hanami specs go green again on my machine.

lib/dry/system/provider/source.rb
alassek: @flash-gordon I'm all in favor of adopting language updates, but the gemspec currently allows 3.0 and up, so I'd have to do it both ways. Is there a benefit to that added complexity in order to use `set_temporary_name`?

flash-gordon: @alassek I have no idea :) It's up to you

timriley: I'm happy if we just leave it for now.

timriley:
@alassek Are you happy with this PR (and https://github.com/hanami/hanami/pull/1446) in their current states?

alassek:
@timriley Yes, thank you for figuring out the `<=` problem (I didn't even know you could do that!)

I'm happy with this.

timriley:
Thanks for all your efforts getting this across the line, @alassek!!

PR #276, opened 2024-07-02
Add eager loading during finalization

### LINKED ISSUE #216: Could container raise error on finalization when component is not resolvable?
## Describe the bug/feature

I ran into an issue where I had defined a constant with an acronym in the name and registered in with `dry-system`. I just needed to define the acronym in the inflector, but I wonder if we should even successfully `finalize!` when a (non-lazy loaded) component is unresolvable. Instead, we could raise an error with a name like `ComponentNotResolvableError`.

Lazy-loaded components won't have their constants defined at finalization, so this would only be for non-lazy loaded components.

## To Reproduce

```shell
git clone git@github.com:cllns/dry-system-unresolvable-component-example.git cllns-dry-system-unresolvable-component-example
cd cllns-dry-system-unresolvable-component-example/
bundle 
bundle exec ruby lib/container.rb
```
You'll get a `NameError` because the class is called `HTML` which isn't recognized by the inflector, (though the `API` class is, so that works fine.)

## Expected behavior

Running `.finalize!` on a container would check to see if constants that correspond to each key that's registered are defined. It shouldn't actually resolve the components, but it should check to see if the components are resolveable (preresolve?)

~I guess the question is: does `dry-system` need to support constants defined  __after__ a container has been `finalize!`'d. 
If it does, then we can wrap `NameError` via implement #147 and that will be helpful enough for users.~ UPDATE: This is complete

I can take a swing at implementing this, if you think it's a worthwhile feature.

cllns:
Fixes #216

### DIFF
--- a/lib/dry/system/container.rb
+++ b/lib/dry/system/container.rb
@@ -314,7 +314,7 @@ module Dry
         # @return [self] frozen container
         #
         # @api public
-        def finalize!(freeze: true, &block)
+        def finalize!(freeze: true, eager_load: false, &block)
           return self if finalized?
 
           configured!
@@ -326,6 +326,10 @@ module Dry
           auto_registrar.finalize!
           manifest_registrar.finalize!
           importer.finalize!
+
+          if eager_load
+            keys.each { |key| resolve(key) }
+          end
 
           @__finalized__ = true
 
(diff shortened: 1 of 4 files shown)


flash-gordon (approved):
Looks good to me! I'm testing it against my apps, I wonder if anything interesting comes up

flash-gordon:
@cllns checks passed, mind adding a changelog entry before merging?

PR #287, opened 2025-07-30
Fix lazy loading of imported components

timriley:
We were checking for an import namespace using a component `root_key`, which is a symbol, whereas the importer was expecting strings. This led to imported components only becoming available on a full boot, not when lazy loading properly (because the check for their namespace in the importer would always return false due to the type mismatch).

This will resolve https://github.com/hanami/hanami/issues/1529.

### DIFF
--- a/lib/dry/system/container.rb
+++ b/lib/dry/system/container.rb
@@ -629,8 +629,8 @@ module Dry
             load_local_component(component)
           elsif manifest_registrar.file_exists?(component)
             manifest_registrar.(component)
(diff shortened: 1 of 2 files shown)


unused:
Instead of adapting the check, how about updating the storage (`Importer@registry`) instead to store the specific type solely `registry[namespace.to_s] = # ... in Importer#register`?

timriley:
@unused thank you for checking this out! :D I agree, that's a much better approach. I've just pushed that up in the last commit here. I'll merge this now and cut a patch release.

PR #295, opened 2026-07-21
ManifestRegistrar#finalize! avoids duplicates

alassek:
This resolves a subtlety introduced by the fix in #286

`Container#finalize!` processes providers first, before moving on to manifest registrations.

If a class injects something defined by a manifest file, and this class is instantiated within a provider, this eager-loads the manifest key before `finalize!` reaches the ManifestRegistrar.

Moving `ManifestRegistrar#call` to use `load` instead of `require` introduced the possibility of double-registering a key.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Break Versioning](https://www.taoensso.com/break-ve
 
 ## [Unreleased]
 
+### Fixed
+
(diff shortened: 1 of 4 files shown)


alassek:
Rubocop failures are not related to this changeset.

timriley:
Thanks for picking this up, @alassek! I pushed a little tweak that I think should make it more efficient — how does that seem to you?

alassek:
@timriley that's a better solution, thanks!

PR #297, opened 2026-09-02
Enable reloading with zeitwerk and use load for provider file loading

afomera:
Out of the Updates to bring code reloading in-house with Zeitwerk this PR does two things, I've split them into their own commits in case we wanted to review separately but it wasn't a ton of code together so leaving it so less PRs to review/manage at once. Let me know if you disagree!

- This updates it to load the provider files with `load` rather than require which only executes the file the first time, load makes it reloadable by default
- Add the enable_reloading plugin for zeitwerk

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,8 @@ and this project adheres to [Break Versioning](https://www.taoensso.com/break-ve
 
 ### Changed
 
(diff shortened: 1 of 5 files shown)


timriley:
Thanks @afomera, this is great! I made a few tiny tweaks and just pushed them directly here. You can see them as separate commits above. Since this is the first change of several that we need for our Hanami slice reloading, I'm just going to go ahead and merge this one now :)

